PR #21, opened 2023-09-04
AMB-1688: Batch Processing Lambda

ChewieCB:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist

### DIFF
--- a/terraform/batch-processing/s3.tf
+++ b/terraform/batch-processing/s3.tf
@@ -1,11 +1,16 @@
+locals {
+    // Flag so we can force delete s3 buckets with items in for pr and shortcode environments only.
+    is_temp = length(regexall("[a-z]{2,4}-?[0-9]{2}", var.environment)) > 0
+}
+
 resource "aws_s3_bucket" "batch_lambda_source_bucket" {
     bucket = "${var.prefix}-batch-lambda-source"
-    force_destroy = true
+    force_destroy = local.is_temp
 }
 
 resource "aws_s3_bucket" "batch_lambda_destination_bucket" {
     bucket = "${var.prefix}-batch-lambda-destination"
-    force_destroy = true
+    force_destroy = local.is_temp
 }
 
 resource "aws_s3_bucket_notification" "batch_processing_source_lambda_trigger" {
--- a/terraform/batch-processing/s3.tf
+++ b/terraform/batch-processing/s3.tf
@@ -1,0 +1,7 @@
+resource "aws_s3_bucket" "source" {
+    bucket = "source"
+}
+
+resource "aws_s3_bucket" "destination" {
+    bucket = "destination"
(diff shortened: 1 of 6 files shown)


terraform/batch-processing/s3.tf
```diff
@@ -0,0 +1,18 @@
+resource "aws_s3_bucket" "batch_lambda_source_bucket" {
+    bucket = "${var.prefix}-batch-lambda-source"
+    force_destroy = true
```
artronics, line 3: This shouldn't be forced if the environment (prefix in this case) is not pr. There is a ternary operator in terraform that you can use. 
The teardown shouldn't happen for any non-pr/shortcode deployment. So, we want to prevent S3 deletion in case it's not empty.

terraform/batch-processing/s3.tf
```diff
@@ -0,0 +1,18 @@
...
+}
+
+resource "aws_s3_bucket" "batch_lambda_destination_bucket" {
+    bucket = "${var.prefix}-batch-lambda-destination"
+    force_destroy = true
```
artronics, line 8: Same as the other s3 bucket comment.

terraform/batch-processing/src/batch_processing.py
```diff
@@ -0,0 +1,27 @@
...
+    with tempfile.TemporaryDirectory() as tmpdir:
+        os.chdir(tmpdir)
+
+        # Write some placeholder bytestring data to a file in the bucket,
+        # so we can test that the lambda writes to the correct output bucket.
+        filename = f"output_report_{time()}.txt"
+        data = (b'Test file to see if the lambda writes to the correct s3 bucket. '
+                b'If our AWS bill skyrockets, this file has been written to the wrong bucket!')
+
+        output_bucket.put_object(Body=data, Key=filename)
```
artronics, line 23: Why we are creating that tmp directory? When we all the actual logic all the data is in memory (it's based on response data). Is it necessary to create a file?

terraform/batch-processing/s3.tf
```diff
@@ -0,0 +1,23 @@
+locals {
+    // Flag so we can force delete s3 buckets with items in for pr and shortcode environments only.
+    is_temp = length(regexall("[a-z]{2,4}-?[0-9]{2}", var.environment)) > 0
```
artronics, line 3: Nice one. It matches both pr and shortcode 🚀 . The only thing is that last {2}. It won't match multiple digits in a pr like pr-123. `[a-z]{2,4}-?[0-9]+` I think will work. I didn't test it though.

almas-liaqat (approved):

PR #47, opened 2023-11-02
 AMB-1769-Convert-MESH-CSV-records-to-FHIR-objects

almas-liaqat:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist


backend/pyutils/csv_to_model.py
```diff
@@ -1,42 +1,60 @@
...
+        "unique_id": row.get("UNIQUE_ID", None),
+        "failure_reasons": reasons,
+    }
+
+    immunisation_failure = ImmunizationErrorModel(**failed_record)
+
+    return immunisation_failure
+
+
+def read_csv_to_model(
```
artronics, line 28: This function shouldn't expect a file. This code should perform under the assumption that it receives the content as string. You can read a file during test and pass it to this function. If this function assumes file then, we can't use it in the actual implementation because the content comes from S3 not reading a file.

lambda_code/tests/test_mesh.py
```diff
@@ -76,21 +83,69 @@ def setUp(self):
     @mock_s3
     def test_write_to_s3(self):
```
artronics, line 84: What's the purpose of this test? If we want to make sure that the content is written back, then we need to mock the `MeshOutputHandler` and assert if the write report has been called.

lambda_code/tests/test_mesh.py
```diff
@@ -76,21 +83,69 @@ def setUp(self):
...
+
+    def test_successful_immunizations_happy_output(self):
+        """Test that the successful immunization record is formatted correctly"""
+        mesh_csv = MeshCsvParser(self.csv_file_path)
+        immunisation_records, _ = mesh_csv.parse()
+        immunisation_record_json = json.dumps(
+            immunisation_records[0].dict(), indent=4, default=str
+        )
+
+        assert json.loads(immunisation_record_json) == json.loads(
```
artronics, line 141: Please use `unittest` assertions. They'll give you diff and they're integrated with the test-runner. Like `self.assertDictEqual()` .

lambda_code/src/old_csv_to_model.py
artronics, line 1: Is this file still necessary? It's called old 😄

lambda_code/src/mesh.py
artronics, line 1: Can you please remove all the `TODO`s if they are not relevant anymore. Only keep the one that still needs addressing. For example you create an Error class which means the one that's in this file is not used anymore.

lambda_code/tests/test_mesh.py
nathanwall: I think this one is yours

artronics (approved):

artronics (approved):

PR #68, opened 2024-01-10
Amb 1850 wire in validations

artronics:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist


lambda_code/tests/test_fhir_repository.py
```diff
@@ -1,21 +1,17 @@
...
 import unittest
 from unittest.mock import MagicMock, patch, ANY
 
 import botocore.exceptions
 from boto3.dynamodb.conditions import Attr, Key
 
-from fhir_repository import ImmunizationRepository, create_table
-from models.errors import ResourceNotFoundError, UnhandledResponseError
-
+sys.path.append(f"{os.path.dirname(os.path.abspath(__file__))}/../src")
```
lee-richardson-nhs, line 11: This should not be necessary. If your editor can't find the paths let me know and I'll have a look.

terraform/main.tf
```diff
@@ -13,7 +13,7 @@ terraform {
     region = "eu-west-2"
     key    = "state"
   }
-  required_version = "1.6.4"
+    required_version = ">= 1.5.0"
```
lee-richardson-nhs, line 16: You sure this works? I used 1.6.4 because it was required the for modules I used, though I'm struggling to find a source for that.
If so then you'll need to change `.terraform-version` as well.

tests/test_validation.py
```diff
@@ -0,0 +1,29 @@
...
+        "level": "aal3",
+        "login_form": {"username": "656005750104"},
+    }
+)
+@pytest.mark.debug
+def test_coarse_validation(nhsd_apim_proxy_url, nhsd_apim_auth_headers):
+    token = nhsd_apim_auth_headers["Authorization"]
+    imms_api = ImmunisationApi(nhsd_apim_proxy_url, token)
+
+    imms = copy.deepcopy(load_example("Immunization/POST-Immunization.json"))
```
lee-richardson-nhs, line 22: Why the copy? It looks like the load gives you your own object. Is there some caching maybe?

lee-richardson-nhs (changes requested):

terraform/main.tf
artronics: This works for me. I think locking this to the newest will cause unnecessary build failures. That `.terraform-version` can be to the latest version (as it is now) and shouldn't cause issues:
>Please note that we don't do semantic version range parsing but use first ever found version as the candidate for minimally required one. It is up to the user to keep the definition reasonable.

tests/test_validation.py
artronics: It caused issues before when I needed to generate objects in a loop. I think it was caused by python caching the object. We don't call it multiple times here but I wanted to be safe.

lambda_code/src/models/errors.py
```diff
@@ -55,6 +55,16 @@ def to_operation_outcome(self) -> OperationOutcome:
...
 
 
+@dataclass
+class CoarseValidationError(ValidationError):
+    """Pre validation error"""
+    message: str
+
+    def to_operation_outcome(self) -> OperationOutcome:
+        return create_operation_outcome(
+            resource_id=str(uuid.uuid4()), severity=Severity.error, code=Code.invalid, diagnostics=self.message)
```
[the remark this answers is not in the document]
lee-richardson-nhs: I can't see where this has been addressed.

lambda_code/src/fhir_service.py
```diff
@@ -1,17 +1,22 @@
-import json
 from typing import Optional
-from fhir.resources.immunization import Immunization
+
+from fhir.resources.R4B.immunization import Immunization
```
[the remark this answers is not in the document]
lee-richardson-nhs: There's a bunch more of these in other files.

lambda_code/src/models/errors.py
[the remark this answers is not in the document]
lee-richardson-nhs: You didn't change the usage of the code.

lee-richardson-nhs (approved):

PR #85, opened 2024-01-30
AMB-1876-Search-endpoint-FHIR-compliance 

almas-liaqat:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist


lambda_code/src/fhir_service.py
```diff
@@ -78,18 +92,25 @@ def delete_immunization(self, imms_id) -> Immunization:
...
-        return FhirList.construct(entry=entries)
+        entries = [
+            BundleEntry(resource=Immunization.parse_obj(handle_s_flag(imms, patient))) for imms in resources
+        ]
+        fhir_bundle = FhirBundle(
+            resourceType="Bundle",
+            type="searchset",  # Set the type to "searchset"
+            entry=entries,
+        )
+        url = f"{get_service_url()}/Immunization?patient={nhs_number}&diseaseType={disease_type}"
```
lee-richardson-nhs, line 111: Shouldn't `patient` be `nhsNumber`?
Or better yet, could the URL be taken from the event so we know for sure it's exactly what was used?

lambda_code/src/fhir_controller.py
```diff
@@ -99,23 +105,67 @@ def delete_immunization(self, aws_event):
...
+                    nhsNumberValue = nhsNumber_list[0]
+
+                diseaseType_list = parsed_body.get(diseaseTypeParam)
+                if diseaseType_list:
+                    diseaseTypeValue = diseaseType_list[0]
+                # Continue processing the request
 
+        params = aws_event["queryStringParameters"]
+        # If nhsNumber was not present in body then it should be in params irresepective of GET/POST
+        if nhsNumberValue is None:
```
lee-richardson-nhs, line 136: The FHIR docs say that putting the same param in the body and the URL is the same as repeating them, not that one is taken over the other. Which makes sense to me, we can't silently ignore one set of params as they could be different.

In which case, looking at the docs, it's an AND. If they're different it can never return anything as an Immunization can't have 2 ids. If they're the same value then it's the same as having just one. 
I.e. `_id=1 AND _id=2` can never return anything but is a valid call.
`id=1 AND id=1` is just `id=1`.

http://hl7.org/fhir/R4B/http.html#search
http://hl7.org/fhir/R4B/search.html#combining

lee-richardson-nhs (changes requested):

lambda_code/src/fhir_service.py
almas-liaqat: Self link in Searchset Bundle can only be a get, thats why i prepared it such a way that irrespective of POST/GET call, it will prepare a self link which will be get. Yeah you are right, that parameter should be nhsNumber

lambda_code/src/fhir_controller.py
```diff
@@ -99,23 +105,80 @@ def delete_immunization(self, aws_event):
             return self.create_response(500, unhandled_error.to_operation_outcome())
 
     def search_immunizations(self, aws_event) -> dict:
+        http_method = aws_event.get("httpMethod")
+        nhsNumberValue = None
+        diseaseTypeValue = None
+        nhsNumberParam = "-nhsNumber"
+        diseaseTypeParam = "-diseaseType"
+        if http_method == "POST":
+            # Check if the Content-Type is application/x-www-form-urlencoded
```
lee-richardson-nhs, line 114: Half of the comments in this section are not accurate to what is being done and the ones that are accurate are rather redundant. I think they can all be removed. Generally code should be self documenting and you'd only need comments for a rare instance of something unavoidably confusing.

lambda_code/src/fhir_controller.py
```diff
@@ -99,23 +105,80 @@ def delete_immunization(self, aws_event):
...
+                # Parse the body of the request
+                body = aws_event["body"]
+                # Decode the base64-encoded body
+                decoded_body = base64.b64decode(body).decode("utf-8")
+                # Parse the decoded body as application/x-www-form-urlencoded
+                parsed_body = urllib.parse.parse_qs(decoded_body)
+                # Access individual form parameters and check if nhsNumber or diseaseType is present
+                nhsNumber_list = parsed_body.get(nhsNumberParam)
+                if nhsNumber_list:
+                    nhsNumberValue = nhsNumber_list[0]
```
lee-richardson-nhs, line 126: This silently ignores multiple values. It needs to be handle the same way as the query string.

lambda_code/src/fhir_controller.py
```diff
@@ -99,23 +105,80 @@ def delete_immunization(self, aws_event):
...
+                return self._create_bad_request(
+                    f"Search Parameter {nhsNumberParam} is mandatory"
+                )
+            else:
+                nhsNumberValue = params[nhsNumberParam]
+        else:
+            #If NhsNumber was present in body as well as in QueryParams, then both values should be same
+            if nhsNumberParam in params:
+                if nhsNumberValue!=params[nhsNumberParam]:
+                    return self._create_bad_request(
```
lee-richardson-nhs, line 145: This is a perfectly valid request by the FHIR docs as far as I can see. In this case it will return no result. Let's leave it for now though as I think we need a separate ticket for sorting out the broken param handling and actual searching because we don't have the details of all the params.

lee-richardson-nhs (changes requested):

lambda_code/src/fhir_controller.py
almas-liaqat: Can you please elaborate this

lambda_code/src/fhir_controller.py
almas-liaqat: We have to return empty Bundle and for that we need to return some self url, what should be the value of those different params in that?
 Should it just be empty or some system generated Incorrect value?

terraform/oas.yaml
```diff
@@ -110,6 +110,32 @@ paths:
             application/fhir+json:
               schema:
                 type: object
+  /event/_search:
```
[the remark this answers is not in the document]
almas-liaqat: That is part of other ticket: [AMB-1871](https://nhsd-jira.digital.nhs.uk/browse/AMB-1871)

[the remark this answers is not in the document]
almas-liaqat: /Immunization is done by Jalal in his Pr, when it will get merge, it will sort it out

lambda_code/src/fhir_controller.py
```diff
@@ -99,30 +116,76 @@ def delete_immunization(self, aws_event):
...
+            content_type = aws_event.get("headers", {}).get("Content-Type")
+            if content_type == "application/x-www-form-urlencoded":
+                body = aws_event["body"]
+                decoded_body = base64.b64decode(body).decode("utf-8")
+                parsed_body = parse_qs(decoded_body)
+                nhs_number_list = parsed_body.get(nhs_number_param)
+                disease_type_list = parsed_body.get(disease_type_param)
+                
+        parsed_query_params = aws_event.get("queryStringParameters", {})
+        # AWS API Gateway doesnot support multiple query string parameters with the same name, so it will not be array
```
lee-richardson-nhs, line 134: API Gateway does support this on another field. They're under `multiValueQueryStringParameters` and contain all the parameters, even singles, so you don't have to combine it with `queryStringParameters` or anything like that.

https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format

almas-liaqat: Yeah but we are not using multiValueQueryStringParameters, thats why dont need it. We are passing our params as queryStringParameters

lee-richardson-nhs: We don't choose how the params are passed, API Gateway always puts them in both fields. If we use `queryStringParameters` in our code then we're losing the other parameters so we need to be using `multiValueQueryStringParameters` in order to make the URL params work the same way as the POST content params.

almas-liaqat: If we want to handle multiple params of same name in request, we should create another ticket for it, its out of scope of this ticket.

lee-richardson-nhs: The changes you've made have handled it now so no need.

lee-richardson-nhs (approved):

PR #86, opened 2024-01-30
AMB-1866 align backend with spec

artronics:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist


lambda_code/tests/test_update_imms.py
```diff
@@ -0,0 +1,53 @@
...
+from fhir_controller import FhirController
+from models.errors import Severity, Code, create_operation_outcome
+
+
+class TestUpdateImmunizations(unittest.TestCase):
+    def setUp(self):
+        self.controller = create_autospec(FhirController)
+
+    def test_update_immunization(self):
+        """it should return 200 if update is successful"""
```
lee-richardson-nhs, line 20: This isn't what's being tested.

terraform/oas.yaml
```diff
@@ -22,7 +22,22 @@ paths:
...
+    put:
+      x-amazon-apigateway-integration:
+        uri: "${update_event.lambda_arn}"
+        payloadFormatVersion: "1.0"
+        passthroughBehavior: "when_no_match"
+        httpMethod: "PUT"
+        timeoutInMillis: 30000
+        type: "AWS_PROXY"
+      responses:
+        '200':
```
lee-richardson-nhs, line 35: Does this need the 201 as well?

tests/test_crud_immunisation_api.py
```diff
@@ -42,23 +42,37 @@ def test_crud_immunization_nhs_login(nhsd_apim_proxy_url, nhsd_apim_auth_headers
...
 
     result = imms_api.get_immunization_by_id(imms_id)
+    res_body = result.json()
 
     assert result.status_code == 200
     assert res_body["id"] == imms_id
 
+    # UPDATE
+    new_imms = copy.deepcopy(imms)
+    new_imms["id"] = imms_id
```
lee-richardson-nhs, line 61: Should probably be `updated_imms` as it's not new if it's for the same id. A new imms would get a 201.

artronics: It is "new" because it's the result of the previous step. But I agree it's confusing. I can't rename it to `updated` because it's not updated yet either lol.
I think I rename it to `update_payload`. This way, we know it's the imms request body that's about to get updated.

terraform/oas.yaml
artronics: Oh I forgot about `oas.yaml`. That's not the only problem. I fixed everything that was missing in there. Removing `content` and also adding `Location` header. Thanks 🚀

tests/test_crud_immunisation_api.py
```diff
@@ -342,7 +348,10 @@ def assert_is_filtered(imms):
         assert all(performer["actor"]["identifier"]["system"] == "https://fhir.nhs.uk/Id/ods-organization-code"
                    for performer in imms["performer"])
 
-    assert_is_not_filtered(created_imms)
+    if is_restricted:
```
lee-richardson-nhs, line 351: Created will never be filtered because it's the original full object which we need to compare against the retrieved one. I think the work to not return the object on create has scuppered this test. Let's talk in dev stand up tomorrow.

lee-richardson-nhs (approved):

PR #100, opened 2024-02-09
AMB-1895 add authorisation for READ

artronics:
## Things to look for
* I moved all the tests from old `tests` directory. Let me know if there is a scenario that's missing
* The most complicated part is `utils/base_test.py`.
* This ticket is not just about refactoring tests but also adding authorisation to the backend. so check `lambda_code/authorization.py` and it's tests. It's also wired in the controller
## Main changes:
* The `tests` directory was removed. Now we have `e2e` directory
* Previous `.env` file should work with the new one
* If you are running tests locally, then you should see all tests passing except `test_deployment.py` which relies on github commit sha.
* Dependency on `pytest-nhsd-apim` and `pytest` were removed. We only use `unittest` instead.
* Every endpoint operation has it's own test file
* Some redundant tests were removed. These are mostly default tests that came with the template repo
* We don't have markers anymore. See the `Makefile` for different category of tests
* The new framework supports "default app" which means we can now run a subset of tests in the int env. As long as the test doesn't use apigee api (like modifying custom attributes) it can be executed in int
* Each operation's happy path tests is executed using all authentication types in a loop. This ensures the authentication is doing it's job. We only do this for happy-path tests because once it's proven a certain auth method is working, then there is no need to repeat it for each test
* Tests are using a set of common config, unitlity, and apis and you can find them in `utils/base_test.py`


e2e/lib/apigee.py
```diff
@@ -0,0 +1,198 @@
...
+class ApigeeProduct:
+    """Data object to create an apigee product"""
+    name: str
+    apiResources: List[str] = field(default_factory=lambda: [])
+    approvalType: str = "auto"
+    attributes: List[dict] = field(default_factory=lambda: [])
+    description: str = "My API product"
+    displayName: str = "My API product"
+    environments: List[str] = field(default_factory=lambda: [ApigeeEnv.INTERNAL_DEV.value])
+    proxies: List[str] = field(default_factory=lambda: ["identity-service-internal-dev",
```
almas-liaqat, line 93: Is it necessary to hard code these values to internal-dev here?

PR #114, opened 2024-02-26
AMB-000 add ref env

artronics:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist

### DIFF
--- a/azure/templates/post-deploy.yml
+++ b/azure/templates/post-deploy.yml
@@ -114,7 +114,10 @@ steps:
       then
         echo "Sandbox env tests"
         marker="not nhsd_apim_authorization and not sandbox_int and not mtls"
-      elif [[ $APIGEE_ENVIRONMENT == "int" ]] || [[ $APIGEE_ENVIRONMENT == "ref" ]]
+      elif [[ $APIGEE_ENVIRONMENT == "ref" ]]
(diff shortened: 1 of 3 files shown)


azure/templates/post-deploy.yml
```diff
@@ -114,7 +114,7 @@ steps:
       then
         echo "Sandbox env tests"
         marker="not nhsd_apim_authorization and not sandbox_int and not mtls"
-      elif [[ $APIGEE_ENVIRONMENT == "int" ]]
+      elif [[ $APIGEE_ENVIRONMENT == "int" ]] || [[ $APIGEE_ENVIRONMENT == "ref" ]]
```
lee-richardson-nhs, line 117: I wouldn't think this can work. Won't this use the `int` credentials for `ref`, which are on different Apigee instances?

artronics: I separate these two. I don't think this will cause anyway though. The int env only runs smoketests so I think those `INT_*` vars are unused. The credentials in the pipeline is the same for all variables.

lee-richardson-nhs (approved):

PR #386, opened 2025-04-15
VED-165-FHIR-TO-FLAT-JSON3-SNOMED

Akol125:
## Summary

* :sparkles: New Feature
Improve Validation Logic for SNOMED-related code and their terms
Improve Validation Logic for Recorded, Expiry and Birthdate
Update Unit test to reflect those changes
Ensure appropriate Error handling are in place with safe return messages to specific fields.

## Reviews Required
* [x] Dev
* [x] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist


nhsdevws (dismissed):

delta_backend/src/ConversionChecker.py
```diff
@@ -85,25 +93,121 @@ def convertData(self, expressionType, expressionRule, fieldName, fieldValue):
...
+
+        self.errorRecords.append({
+            "code": code,
+            "field": fieldName,
+            "value": fieldValue,
+            "message": message
+        })
 
-    # Convert ISO date string to a specific format (e.g. YYYYMMDD)
     def _convertToDate(self, expressionRule, fieldName, fieldValue, summarise, report_unexpected_exception):
```
robertnovac1, line 116: Maybe one for later, but this seems over-engineered. As in can't we have a desired state of how we accept date times and reject everything else like partial dates, dates with missing timezone, etc? Something like

```
def validate_date(date_str):
    try:
        # Parse string to datetime
        dt = datetime.fromisoformat(date_str)

        # Check if it includes timezone info
        if dt.tzinfo is None:
            return False, "Missing timezone information"

        # Convert to UTC to safely compare
        now = datetime.now(timezone.utc)
        dt_utc = dt.astimezone(timezone.utc)

        # Check if date is in the future
        if dt_utc > now:
            return False, "Date is in the future"

        # Check if year is >= 2020
        if dt_utc.year < 2020:
            return False, "Date must be from 2020 onward"

        return True, dt

    except ValueError:
        return False, "Invalid date format – must be ISO 8601 (e.g. YYYY-MM-DDTHH:MM:SS±HH:MM)"
```

delta_backend/src/ConversionChecker.py
```diff
@@ -266,3 +371,45 @@ def _convertToOnlyIfTo(self, expressionRule, fieldName, fieldValue, summarise, r
...
+        """
+        try:
+            if not fieldValue:
+                return fieldValue
+            if not isinstance(fieldValue, str) or not fieldValue.isdigit():
+                raise ValueError(f"Invalid SNOMED code: {fieldValue}")
+            return fieldValue
+        except Exception as e:
+            if report_unexpected_exception:
+                message = ExceptionMessages.MESSAGES[ExceptionMessages.UNEXPECTED_EXCEPTION] % (e.__class__.__name__, e)
```
robertnovac1, line 388: Same as in the comment before, it would be more readable if you would use f-string rather than %.

delta_backend/src/Converter.py
```diff
@@ -28,6 +28,24 @@ class Converter:
...
         self.SchemaFile = ConversionLayout.ConvertLayout
+        # self.ErrorRecords = []
+    
+    # Utility logs tailored to conveter class errors
+    def _log_error(self,e,code=ExceptionMessages.UNEXPECTED_EXCEPTION): 
+        message = str(e)  # if a simple string message was passed 
+    
+        if any(existing.get("message") == message for existing in ErrorRecords): 
+            return
+
```
robertnovac1, line 39: The indentation for error_obj is off a bit

delta_backend/tests/test_convert_to_flat_json.py
robertnovac1, line 1: This file has a couple of formatting issues in the space between methods

delta_backend/tests/test_convert_to_flat_json.py
```diff
@@ -65,12 +94,12 @@ def setUp(self):
                 },
             ],
         )
-
+
```
robertnovac1, line 97: These things should be fixed

delta_backend/tests/test_convert_to_flat_json.py
```diff
@@ -327,25 +386,76 @@ def test_convert_to_nhs_number(self, MockLookUpData):
         invalid_nhs_number = "1234567890243"
         result = checker._convertToNHSNumber("NHSNUMBER","fieldName", invalid_nhs_number, False, True)
         self.assertEqual(result, "", "Invalid NHS number should return empty string")
-
+   
     @patch("ConversionChecker.LookUpData")
     def test_convert_to_date(self, MockLookUpData):
```
robertnovac1, line 391: Probably needs splitting up to test the various scenarios. Unit tests should cover an isolated piece of code. If this fails, it's difficult to understand which bit failed. This brings me back to what I said about it being over-engineered. It should accept only 1 format and reject anything else. This is probably why you got the issues with the unit test coverage from sonarqube.

delta_backend/src/ConversionChecker.py
```diff
@@ -85,25 +93,121 @@ def convertData(self, expressionType, expressionRule, fieldName, fieldValue):
                     expressionRule, fieldName, fieldValue, self.summarise, self.report_unexpected_exception
                 )
             case _:
-                return "Schema expression not found! Check your expression type : " + expressionType
+               return "Schema expression not found! Check your expression type : " + expressionType
```
mfjarvis, line 96: Should this call `_log_error` instead of returning an error message?

delta_backend/src/ConversionChecker.py
```diff
@@ -148,14 +252,17 @@ def _convertToDateTime(self, expressionRule, fieldName, fieldValue, summarise, r
...
         try:
-            if len(str(fieldValue)) > 0:
+            if isinstance(fieldValue, (int, float)):
+                return ""  # Reject plain numbers
+            if isinstance(fieldValue, str) and fieldValue.strip():
                 return fieldValue
             return ""
         except Exception as e:
             if report_unexpected_exception:
                 message = ExceptionMessages.MESSAGES[ExceptionMessages.UNEXPECTED_EXCEPTION] % (e.__class__.__name__, e)
```
mfjarvis, line 262: Should this call `_log_error` instead of returning the error message?

delta_backend/src/Converter.py
```diff
@@ -74,10 +92,8 @@ def runConversion(self, json_data, summarise=False, report_unexpected_exception=
             dataParser = self._getFHIRParser(self.FHIRData)
         except Exception as e:
             if report_unexpected_exception:
-                message = "FHIR Parser Unexpected exception [%s]: %s" % (e.__class__.__name__, e)
-                p = {"code": 0, "message": message}
-                ErrorRecords.append(p)
-                return p
+                error = self._log_error("FHIR Parser Unexpected exception [%s]: %s" % (e.__class__.__name__, e),code=0)
```
robertnovac1, line 95: Use f string for readability

delta_backend/src/ConversionChecker.py
mfjarvis: There are a few other similar pieces of code in this file

robertnovac1 (changes requested):

delta_backend/src/ConversionChecker.py
```diff
@@ -85,25 +93,121 @@ def convertData(self, expressionType, expressionRule, fieldName, fieldValue):
                     expressionRule, fieldName, fieldValue, self.summarise, self.report_unexpected_exception
                 )
             case _:
-                return "Schema expression not found! Check your expression type : " + expressionType
+               return "Schema expression not found! Check your expression type : " + expressionType
+    
+    # Utility function for logging errors
+    def _log_error(self, fieldName, fieldValue, e, code=ExceptionMessages.RECORD_CHECK_FAILED):
```
mfjarvis, line 99: Do we always check `self.report_unexpected_exception` before calling `_log_error`? If so, could we check it once here instead?

delta_backend/src/ConversionChecker.py
```diff
@@ -85,25 +93,121 @@ def convertData(self, expressionType, expressionRule, fieldName, fieldValue):
...
+        self.errorRecords.append({
+            "code": code,
+            "field": fieldName,
+            "value": fieldValue,
+            "message": message
+        })
 
-    # Convert ISO date string to a specific format (e.g. YYYYMMDD)
     def _convertToDate(self, expressionRule, fieldName, fieldValue, summarise, report_unexpected_exception):
+        print("FIELD:", fieldName)
```
mfjarvis, line 117: Debug logging?

delta_backend/src/ConversionChecker.py
mfjarvis, line 388: Could this be a call to `_log_error` instead?

delta_backend/src/ConversionChecker.py
```diff
@@ -266,3 +371,45 @@ def _convertToOnlyIfTo(self, expressionRule, fieldName, fieldValue, summarise, r
...
+            if isinstance(fieldValue, bool):
+                return fieldValue
+
+            if isinstance(fieldValue, str):
+                lowered = fieldValue.strip().lower()
+                if lowered == "true":
+                    return True
+                elif lowered == "false":
+                    return False
+            return ""
```
mfjarvis, line 408: Should it not be an error if we receive something other than a `bool` or `str`, or if we receive an invalid string?

delta_backend/tests/sample_data/fhir_sample.json
```diff
@@ -46,7 +46,7 @@
         "coding": [
           {
             "system": "http://snomed.info/sct",
-            "code": "1324681000000101",
+            "code": "13246810000001git",
```
mfjarvis, line 49: Is this a typo or a deliberate invalid value?

Akol125: This is a bad sample data to test the conversion. its deliberate

delta_backend/src/ConversionChecker.py
Akol125: dynamic messages at run-time, f-string won't work for it

Akol125: yes it can, will change it

delta_backend/src/ConversionChecker.py
```diff
@@ -85,25 +93,121 @@ def convertData(self, expressionType, expressionRule, fieldName, fieldValue):
...
+                    self._log_error(fieldName, fieldValue, "Date must be in YYYYMMDD format")
+                return ""
+
         try:
-            dt = datetime.fromisoformat(fieldValue)
-            format_str = expressionRule.replace("format:", "")
+            dt = datetime.strptime(fieldValue, format_str)
+
+            # Reject future dates if the field is BirthDate
+            if fieldName in "contained|#:Patient|birthDate":
```
mfjarvis, line 198: I don't love referring to field names in the conversion logic - can this be controlled by the schema instead? Same question for `recorded` above

nhsdevws (commented):
sonarcloud is failing with firehose mocking issues

robertnovac1 (dismissed):

delta_backend/.coverage
mfjarvis, line 1: Should this be in version control?

delta_backend/data.json
mfjarvis, line 1: Should this be in version control?

delta_backend/src/ConversionChecker.py
```diff
@@ -85,25 +93,120 @@ def convertData(self, expressionType, expressionRule, fieldName, fieldValue):
...
+
+    # Utility function for logging errors
+    def _log_error(self, fieldName, fieldValue, e, code=ExceptionMessages.RECORD_CHECK_FAILED):
+        if isinstance(e, Exception):
+            message = ExceptionMessages.MESSAGES[ExceptionMessages.UNEXPECTED_EXCEPTION] % (e.__class__.__name__, str(e))
+        else:
+            message = str(e)  # if a simple string message was passed 
+
+        # Ensure message-level deduplication
+        if any(existing.get("message") == message for existing in self.errorRecords):
```
mfjarvis, line 106: Does this mean if we have the same type of error for two different fields, then we'll only report one of the errors? Maybe we should deduplicate on the entire error object, or not deduplicate at all?

mfjarvis: How would we actually end up with duplicate error messages?

delta_backend/src/ConversionChecker.py
```diff
@@ -85,25 +93,120 @@ def convertData(self, expressionType, expressionRule, fieldName, fieldValue):
...
+                self._log_error(fieldName, fieldValue, "Partial date not accepted")
             return ""
+
+        # Handle only the recorded field with extended ISO + timezone support
+        if fieldName == "recorded":
+            # Accept "YYYY-MM-DD" and return as is
+            if re.match(r"^\d{4}-\d{2}-\d{2}$", fieldValue):
+                try:
+                    dt = datetime.strptime(fieldValue, "%Y-%m-%d")
+                    if dt.date() > datetime.now(ZoneInfo("UTC")).date():
```
mfjarvis, line 140: Definitely an edge case, but what if we receive a vaccination event at say 00:30 BST? Then `now` in UTC would be 23:30 on the previous day and therefore the recorded date would appear to be in the future. (I think? 😅)

delta_backend/src/ConversionChecker.py
```diff
@@ -223,17 +328,15 @@ def _convertToDose(self, expressionRule, fieldName, fieldValue, summarise, repor
             return fieldValue
         return "" 
 
-    # Change to Lookup
+    # Change to Lookup (loads expected data as is but if empty use lookup extraction to populate value)
     def _convertToLookUp(self, expressionRule, fieldName, fieldValue, summarise, report_unexpected_exception):
+        if isinstance(fieldValue, str) and any(char.isalpha() for char in fieldValue) and not fieldValue.isdigit():
```
mfjarvis, line 333: What's the reason for performing this specific check and not just looking for any non-empty string?

delta_backend/src/ConversionLayout.py
```diff
@@ -140,15 +140,15 @@
...
+        "expressionRule": "%Y%m%d"
       }
     },
     {
       "fieldNameFHIR": "primarySource",
       "fieldNameFlat": "PRIMARY_SOURCE",
       "expression": {
         "expressionName": "Not Empty",
-        "expressionType": "NOTEMPTY",
+        "expressionType": "BOOLEAN",
```
mfjarvis, line 151: A few `expressionName` and `expressionType` mismatches in this file.

Maybe a question for later, but why do we even have both? I'm assuming only the `expressionType` matters?

delta_backend/src/Converter.py
```diff
@@ -28,6 +27,24 @@ class Converter:
...
+    
+        if any(existing.get("message") == message for existing in self.ErrorRecords): 
+            return
+        
+        error_obj = {
+            "code": code,
+            "message": message
+        }
+
+        # Ensure message-level deduplication
```
mfjarvis, line 44: Can remove this commented out code

delta_backend/src/Converter.py
```diff
@@ -56,7 +73,7 @@ def _convertData(self, ConversionValidate, expression, dataParser, json_data):
         except Exception as e:
             message = "Data get value Unexpected exception [%s]: %s" % (e.__class__.__name__, e)
             p = {"code": ExceptionMessages.PARSING_ERROR, "message": message}
-            ErrorRecords.append(p)
+            self.ErrorRecords.append(p)
```
mfjarvis, line 76: Should this use `_log_error`?

mfjarvis: And should it check `report_unexpected_exception`?

mfjarvis: Ah did we say we'd fix those places in another PR?

delta_backend/tests/test_convert_to_flat_json.py
```diff
@@ -334,18 +392,108 @@ def test_convert_to_date(self, MockLookUpData):
...
+        result = checker._convertToDate("%Y%m%d", "fieldName", 12345678, False, True)
+        self.assertEqual(result, "")
+
+        # 6. Future date for birthDate (should trigger "Date cannot be in the future")
+        future_date = (datetime.now() + timedelta(days=365)).strftime("%Y%m%d")
+        result = checker._convertToDate("%Y%m%d", "contained|#:Patient|birthDate", future_date, False, True)
+        self.assertEqual(result, "")
+        
+        # # 7. Valid recorded date in the future (should also be rejected)
+        # result = checker._convertToDate("%Y%m%d", "recorded", future_date, False, True)
```
mfjarvis, line 421: Why is this one commented out?

delta_backend/tests/test_convert_to_flat_json.py
```diff
@@ -334,18 +392,108 @@ def test_convert_to_date(self, MockLookUpData):
...
+        # self.assertEqual(result, "")
+
+        # 8. Empty string
+        result = checker._convertToDate("%Y%m%d", "fieldName", "", False, True)
+        self.assertEqual(result, "")
+
+        # 9. Valid recorded date with timezone
+        valid_recorded = datetime.now(ZoneInfo("UTC")).replace(microsecond=0).isoformat()
+        result = checker._convertToDate("format:%Y-%m-%d", "recorded", valid_recorded, False, True)
+        self.assertTrue(result.startswith(datetime.now(ZoneInfo("UTC")).strftime("%Y%m%dT%H")))
```
mfjarvis, line 431: I think it might be useful to use literal values in some of these test cases rather than generating them - it makes it clearer to read what we're expecting. E.g.:
```suggestion
        valid_recorded = "2025-04-24T09:41:41+00:00"
        result = checker._convertToDate("format:%Y-%m-%d", "recorded", valid_recorded, False, True)
        self.assertTrue(result.startswith("20250424T09"))
```

delta_backend/tests/test_convert_to_flat_json.py
```diff
@@ -334,18 +392,108 @@ def test_convert_to_date(self, MockLookUpData):
...
+        # result = checker._convertToDate("%Y%m%d", "recorded", future_date, False, True)
+        # self.assertEqual(result, "")
+
+        # 8. Empty string
+        result = checker._convertToDate("%Y%m%d", "fieldName", "", False, True)
+        self.assertEqual(result, "")
+
+        # 9. Valid recorded date with timezone
+        valid_recorded = datetime.now(ZoneInfo("UTC")).replace(microsecond=0).isoformat()
+        result = checker._convertToDate("format:%Y-%m-%d", "recorded", valid_recorded, False, True)
```
mfjarvis, line 430: Maybe I'm misunderstanding here, but why are we outputting something with a time when we've asked the converter to output `%Y-%m-%d` format?

robertnovac1 (approved):

PR #396, opened 2025-04-25
VED-227- FHIR-FLAT-JSON

Akol125:
## Summary
- EXPIRY_DATE, PERSON_DOB, and RECORDED_DATE fields are now strictly formatted as YYYYMMDD
- Milliseconds are removed upfront from date-time fields before parsing.
- Refactored unit tests for better error handling and validation

## Reviews Required
* [x] Dev
* [x] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist


delta_backend/tests/test_convert_to_flat_json.py
```diff
@@ -397,60 +397,19 @@ def test_convert_to_date(self, MockLookUpData):
         result = checker._convertToDate("%Y%m%d", "fieldName", 12345678, False, True)
         self.assertEqual(result, "")
 
-        # 6. Future date for birthDate (should trigger "Date cannot be in the future")
```
nhsdevws, line 400: Test for future date should not be removed. 

Instead, test for future should be included and check that it passes

delta_backend/tests/test_convert_to_flat_json.py
```diff
@@ -397,60 +397,19 @@ def test_convert_to_date(self, MockLookUpData):
...
-        expected = "20230415T103000"
-        self.assertTrue(result.endswith("00"),f"Expected prefix {expected}, got {result!r}")
-
-        # 14. Recorded timestamp without tzinfo in the future → rejected
-        future_naive = "2099-12-31T23:59:59"
-        result = checker._convertToDate(format,"recorded",future_naive,False,True)
-        self.assertEqual(result, "")
-
-        # 15 Validate all error logs of various responses
+        # 7 Validate all error logs of various responses
```
nhsdevws, line 404: A series of tests that test one thing can be written as a data array. Makes it easier to read and spot errors.
Optional change.

```
test_data = [
        ["%Y%m%d", "2022-01-01", "20220101", "Valid full date"],
        ["%Y%m%d", "2022-01-01T12:00:00+00:0", "20220101", "Full ISO date"],
        ["%Y%m%d", "invalid_date", "", "Invalid string date format"],
        ["%Y%m%d", None, "", "None input"],
        ["%Y%m%d", 12345678, "", "Not a string input"],
        ["%Y%m%d", "", "", "Empty string"],
        ["format:%Y-%m-%d", "not-a-date", "", "Invalid date for given format"],
    ]

    # Iterate over the test data and run assertions
    for format, value, expected, description in test_data:
        with self.subTest(description):
            result = checker._convertToDate(format, "fieldName", value, False, True)
            self.assertEqual(result, expected, f"Failed for: {description}")
```

nhsdevws (commented):
Put future date back in the tests.

delta_backend/tests/test_convert_to_flat_json.py
Akol125: Yeah, Paul specifically told us to remove it, that we don't need it, plus he mentioned it would be validated from the API/ Batch processor side

nhsdevws (approved):

delta_backend/tests/sample_data/fhir_sample.json
```diff
@@ -153,4 +153,4 @@
       "doseNumberPositiveInt": 2
     }
   ]
-}
+}
```
robertnovac1, line 156: Spacing

delta_backend/tests/test_convert_to_flat_json.py
```diff
@@ -315,7 +315,7 @@ def test_conversion_exceptions(self, mock_get_key_value, mock_get_conversions):
     def test_log_error(self, MockLookUpData):
         # Instantiate ConversionChecker
         checker = ConversionChecker(dataParser=None, summarise=False, report_unexpected_exception=True)
-
```
robertnovac1, line 318: Space

robertnovac1 (dismissed):

delta_backend/src/ConversionChecker.py
```diff
@@ -110,99 +110,35 @@ def _log_error(self, fieldName, fieldValue, e, code=ExceptionMessages.RECORD_CHE
...
                     return ""
-
-            return dt.strftime(format_str)
-        except ValueError as e:
-            if report_unexpected_exception:
-                self._log_error(fieldName, fieldValue, e)
-            return ""
+            try:
+                # Converts raw fieldvalue without delimiters to a date-time object
+                dt = datetime.strptime(fieldValue, expressionRule)
```
mfjarvis, line 135: This seems a bit odd to me - we're parsing a date in a particular format and then serialising it in the same format. So this method is essentially for validating the date format rather than converting it? I'd expect us to parse the date / datetime in a very lenient way, such as `datetime.fromisoformat()`, then serialise using the format string. I'll check with Paul though before we change anything

delta_backend/src/ConversionChecker.py
```diff
@@ -110,99 +110,35 @@ def _log_error(self, fieldName, fieldValue, e, code=ExceptionMessages.RECORD_CHE
...
 
-        try:
-            dt = datetime.strptime(fieldValue, format_str)
-
-            # Reject future dates if the field is BirthDate
-            if fieldName in "contained|#:Patient|birthDate":
-                today_utc = datetime.now(ZoneInfo("UTC")).date()
-                if dt.date() > today_utc:
+            # 2. Use Expression Rule Format to parse the date, remove dashes and slashes
+            if expressionRule == "%Y%m%d":
```
mfjarvis, line 126: As below, it seems odd that we only handle `YYYYMMDD`, `YYYY-MM-DD` or `YYYY-MM-DDT...` here. Why not just parse any valid ISO formatted date / datetime? Will check with Paul

delta_backend/src/ConversionChecker.py
```diff
@@ -246,6 +182,8 @@ def _convertToDateTime(self, expressionRule, fieldName, fieldValue, summarise, r
 
     # Not Empty Validate - Returns exactly what is in the extracted fields no parsing or logic needed
     def _convertToNotEmpty(self, expressionRule, fieldName, fieldValue, summarise, report_unexpected_exception):
+        if not fieldValue:
```
mfjarvis, line 185: Why has this been added? Maybe I'm missing the point of `NOTEMPTY`, but I'd expect `None` to be an error (which would already be picked up by the logic below)

delta_backend/src/ConversionChecker.py
mfjarvis: E.g. what about partial dates?

delta_backend/src/ConversionChecker.py
Akol125: thanks for this Matt, a much simpler path actually.

delta_backend/src/ConversionChecker.py
Akol125: Sorted

delta_backend/src/ConversionChecker.py
Akol125: I was kinda thinking that if its empty the overall data from other data sources is also empty which means its an empty data, hence not exactly an error. what would you suggest?

mfjarvis: Might just be a naming issue - `convertToNotEmpty` makes me think it will never return an empty string, but I might be wrong. Will check with Paul what his intention was. Don't think this is a blocker for getting this MR merged

robertnovac1 (approved):

PR #398, opened 2025-04-30
VED-222-FHIR-FLAT-JSON

Akol125:
## Summary
* Routine Change
  - Output now consistently uses YYYYMMDDThhmmsszz format with GMT/BST support
  - Ensured timezone info is applied if missing
  - Removed fractional seconds in final output
 
  - Upgraded Python from 3.10 to 3.11 (Docker, poetry & venv)

## Reviews Required
* [x] Dev
* [x] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist


delta_backend/src/ConversionChecker.py
```diff
@@ -136,40 +136,36 @@ def _convertToDateTime(self, expressionRule, fieldName, fieldValue, summarise, r
...
+        if offset is not None and offset not in allowed_offsets:
+            if report_unexpected_exception:
+                self._log_error(fieldName, fieldValue, "Unsupported Format or offset")
+            return "" 
 
         # Convert to UTC
-        dt_utc = dt.astimezone(ZoneInfo("UTC")).replace(microsecond=0)
-
-        format_str = expressionRule.replace("format:", "")
+        dt_utc = dt.replace(microsecond=0)
```
mfjarvis, line 162: The code is doing the right thing, but the comment and variable name are wrong - this isn't (necessarily) a UTC datetime

delta_backend/src/ConversionChecker.py
```diff
@@ -136,40 +136,36 @@ def _convertToDateTime(self, expressionRule, fieldName, fieldValue, summarise, r
...
 
-        if format_str == "csv-utc":
+        if expressionRule == "csv-utc":
             formatted = dt_utc.strftime("%Y%m%dT%H%M%S%z")
             return formatted.replace("+0000", "00").replace("+0100", "01")
 
-        return dt_utc.strftime(format_str)
+        return dt_utc.strftime("%Y%m%dT%H%M%S%z")
```
mfjarvis, line 168: As above, I don't think we'll ever hit this case as we only use an `expressionRule` of `csv-utc`

delta_backend/src/ConversionLayout.py
```diff
@@ -68,7 +68,7 @@
       "expression": {
         "expressionName": "Date Convert",
         "expressionType": "DATETIME",
-        "expressionRule": "format:csv-utc"
+        "expressionRule": "csv-utc"
```
mfjarvis, line 71: I think `csv-utc` is now a misnomer as we can also output BST datetimes. I'd suggest either renaming this or just using a blank string (as we only use a single datetime format anyway)

delta_backend/src/ConversionChecker.py
Akol125: that's correct, didn't really change it from previous code. Will refactor

delta_backend/pyproject.toml
```diff
@@ -9,7 +9,7 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = "~3.10"
+python = "~3.11"
```
mfjarvis, line 12: As we've now upgraded Python, can we remove the regex-based microsecond stripping in the `_convertToDate` function?

delta_backend/src/ConversionChecker.py
```diff
@@ -136,40 +136,36 @@ def _convertToDateTime(self, expressionRule, fieldName, fieldValue, summarise, r
...
-        if re.match(r"^\d{4}(-\d{2})?$", fieldValue):
-            raise RecordError(
-                ExceptionMessages.RECORD_CHECK_FAILED,
-                f"{fieldName} rejected: partial datetime not accepted.",
-                f"Invalid partial datetime: {fieldValue}",
-            )
         try:
-            dt = datetime.fromisoformat(fieldValue)
-        except ValueError:
+            if not expressionRule:
```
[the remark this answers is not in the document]
Akol125: not exactly - so the logic is 
YYYY-MM-DD would output YYYYMMDDTThhmmsszz
YYYY-MM-DD-TThh:mm:ss+zz would output YYYYMMDDTThhmmsszz and 
YYYY-MM-DD-TThh:mm:ss would output YYYYMMDDTThhmmsszz

so csv-utc (which will be renamed)  will be = YYYY-MM-DD-TThh:mm:ss+zz
any other format will hit the - if not expressionRule logic (aka - csv-utc (for now))

delta_backend/pyproject.toml
Akol125: sure thing

delta_backend/src/ConversionLayout.py
Akol125: renamed to fhir-date

delta_backend/src/ConversionChecker.py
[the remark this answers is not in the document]
Akol125: sorted

delta_backend/src/ConversionChecker.py
Akol125: sorted

delta_backend/src/ConversionLayout.py
Akol125: sorted

delta_backend/pyproject.toml
Akol125: sorted

mfjarvis (approved):

PR #402, opened 2025-05-02
VED-241 SNOMED converter logic

robertnovac1:
## Summary
* Routine Change
* :sparkles: New Feature

Included the following changes: 
- Fixed search for VACCINE_PRODUCT_CODE so it retrieves record with correct system id 
- Added expression logic to check SNOMED validity only for VACCINE_PRODUCT_CODE. Note this is configurable and it's turned off at the moment.
- Fixed scanning of array to handle URLs
- Moved existing method of getting sample imms request to utils_for_converter_tests.py

## Reviews Required
* [x] Dev

## Review Checklist


delta_backend/src/ConversionChecker.py
```diff
@@ -289,7 +289,7 @@ def _convertToSnomed(self, expressionRule, fieldName, fieldValue, summarise, rep
         """
         try:
             if not fieldValue:
-                return fieldValue
+                return "null"
```
mfjarvis, line 292: Why `"null"` and not `""`? I'd be wary of populating something other than a valid SNOMED code - not sure how this might affect downstream systems

robertnovac1: Yeah, it's because of this requirement about snomed codes:

VACCINE_PRODUCT_CODE
Select first code from coding where system=http://snomed.info/sct and code passes SNOMED regex that it matches SNOMED CT code format else populate null

SITE_OF_VACCINATION_CODE
If exists, select first code from coding where system=http://snomed.info/sct else populate null

robertnovac1: Updated to return default empty value as discussed.

delta_backend/src/FHIRParser.py
```diff
@@ -21,18 +41,21 @@ def _scanValuesForMatch(self, parent, matchValue):
...
+                for key, value in parent[index].items():
+                    if (
+                        (value == fieldList[1] or key == fieldList[1])
+                        and self._validate_expression_rule(expression_type, expression_rule, parent[index])
+                    ):
                         nodeId = index
                         break
                     else:
-                        if self._scanValuesForMatch(parent[index][key], fieldList[1]):
+                        if self._scanValuesForMatch(value, fieldList[1]):
```
mfjarvis, line 58: This branch is confusing me a bit. If we don't find a match in the keys / values of the object we're searching, we then start searching through nested values too? So should we be calling `_validate_expression_rule` here?

Kind of irrelevant because I don't think this functionality to search nested values is ever used. The whole thing feels a bit over-engineered and inefficient to me 😅

mfjarvis: Anyway, it's not new code, so I'm happy to ignore for now. We can look to tidy this whole thing up once we're done with the essential go live work.

mfjarvis (approved):

PR #416, opened 2025-05-16
VED-282-284-226 Delta refactoring extractor

robertnovac1:
## Summary
* Routine Change
* :sparkles: New Feature

Major overhaul of the delta JSON extraction logic. 
Removed the following: 
- FHIRParser.py
- SchemaParser.py
- ConversionChecker.py

Added:
- Refactored the converter to use tailored functionality to extract each field from the immunization event received by the delta. 
- Unit tests for all extraction logic 
- Updated readme.md
- Simplified the conversion layout.

## Reviews Required
* [x] Dev

## Review Checklist


delta_backend/src/conversion_checker.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def __repr__(self):
+        return repr((self.code, self.message, self.details))
+
+# ---------------------------------------------------------------------------------------------------------
+# main conversion checker
+# Conversion engine for expression-based field transformation
+class ConversionChecker:
+    # checker settings
+    summarise = False
```
mfjarvis, line 28: I believe `summarise` and `report_unexpected_exception` can be removed now

delta_backend/src/delta.py
```diff
@@ -72,9 +72,9 @@ def handler(event, context):
                     operation = new_image["Operation"]["S"]
                     action_flag = ActionFlag.CREATE if operation == Operation.CREATE else operation
                     resource_json = json.loads(new_image["Resource"]["S"])
-                    FHIRConverter = Converter(json.dumps(resource_json))
-                    flat_json = FHIRConverter.runConversion(resource_json)  # Get the flat JSON
-                    error_records = FHIRConverter.getErrorRecords()
+                    FHIRConverter = Converter(json.dumps(resource_json), action_flag = action_flag)
```
mfjarvis, line 75: Assuming this is going to be cleaned up, but why are we parsing, then stringifying, then parsing again?

delta_backend/src/fhir_parser.py
```diff
@@ -93,11 +93,10 @@ def _scanForValue(self, FHIRFields, expression_type, expression_rule: str = ""):
         return rootfield
 
     # get the value for a key
-    def getKeyValue(self, fieldName, flatFieldName, expression_type: str = "", expression_rule = ""):
+    def get_key_value(self, fieldName, flatFieldName, expression_type: str = "", expression_rule = ""):
```
mfjarvis, line 96: I think we can remove this entire file now 🙌

delta_backend/src/schema_parser.py
```diff
@@ -0,0 +1,15 @@
+# Schema Parser
+# Moved from file loading to JSON string better for elasticache
+
+
+class SchemaParser:
```
mfjarvis, line 5: I don't think this needs to be a class. Could just do `conversion_layout["conversions"]` where we're currently calling `get_conversions`

delta_backend/tests/sample_data/ConversionLayout.json
```diff
@@ -234,7 +234,7 @@
       "fieldNameFlat": "SITE_OF_VACCINATION_TERM",
       "expression": {
         "expressionName": "Look Up",
-        "expressionType": "LOOKUP",
+        "expressionType": "NORMAL",
```
mfjarvis, line 237: Should we just use the real `conversion_layout.py` in the tests?

delta_backend/src/delta.py
robertnovac1: Yeah that looks strange, I've fixed this

delta_backend/src/fhir_parser.py
robertnovac1: fhir_parser has been removed

delta_backend/src/schema_parser.py
robertnovac1: schema_parser has been removed

delta_backend/tests/sample_data/ConversionLayout.json
robertnovac1: This file has been removed

delta_backend/src/converter.py
```diff
@@ -0,0 +1,62 @@
+# Main validation engine
+import exception_messages
+from conversion_layout import ConversionLayout
+from extractor import Extractor
+
+class Converter:
+
+    def __init__(self, fhir_data, action_flag = "UPDATE", report_unexpected_exception=True):
```
mfjarvis, line 8: Can use `ActionFlag.UPDATE`

delta_backend/src/converter.py
```diff
@@ -0,0 +1,62 @@
...
+        self.report_unexpected_exception = report_unexpected_exception
+               
+        try:
+            if not fhir_data:
+                raise ValueError("FHIR data is required for initialization.")
+        
+            self.extractor = Extractor(fhir_data, self.report_unexpected_exception)
+            self.conversion_layout = ConversionLayout(self.extractor)           
+        except Exception as e:
+            if report_unexpected_exception:
```
mfjarvis, line 21: Do we need to raise here? Otherwise we'll carry on trying to convert with an uninitialised `Converter`

delta_backend/src/delta.py
```diff
@@ -72,9 +72,9 @@ def handler(event, context):
                     operation = new_image["Operation"]["S"]
                     action_flag = ActionFlag.CREATE if operation == Operation.CREATE else operation
                     resource_json = json.loads(new_image["Resource"]["S"])
-                    FHIRConverter = Converter(json.dumps(resource_json))
-                    flat_json = FHIRConverter.runConversion(resource_json)  # Get the flat JSON
-                    error_records = FHIRConverter.getErrorRecords()
+                    FHIRConverter = Converter(resource_json, action_flag = action_flag)
+                    flat_json = FHIRConverter.run_conversion()  # Get the flat JSON
+                    error_records = FHIRConverter.get_error_records()
                     flat_json["ACTION_FLAG"] = action_flag
```
mfjarvis, line 78: Can remove this line

delta_backend/src/converter.py
robertnovac1: True, I wanted to remove this

delta_backend/src/delta.py
robertnovac1: Ahh nice, the merge from master put that line back

delta_backend/src/converter.py
robertnovac1: Fixed

delta_backend/tests/readme.md
```diff
@@ -9,15 +9,15 @@ The basic layout for conversion is below:
 FHIR_data_folder = Path("C:/Source Code/Vaccs to JSON Converter/FHIR-data")
 FHIRFilePath = FHIR_data_folder / "vaccination7.json"
 
-FHIRConverter = Converter(FHIRFilePath)
-FlatFile = FHIRConverter.runConversion(False, True)
+fhir_converter = Converter(FHIRFilePath)
```
mfjarvis, line 12: This file is very outdated; shall we just remove it? I think the conversion is pretty self-explanatory following your rewrite

robertnovac1: Removed

Akol125 (commented):
It's actually more straightforward, rather than use switch cases and expressionRules. Extractor is called in layout and it directly handles conversion logic, nice. The default postcode with zz, is that how you want to handle the default address if none, because it gives an impression that obfuscation was done on the address, which might be misleading, what do you think?

delta_backend/src/conversion_layout.py
```diff
@@ -0,0 +1,146 @@
+from extractor import Extractor
+
+class ConversionLayout:
+    def __init__(self, extractor: Extractor):        
+        self.extractor = extractor
```
nhsdevws, line 5: If you use a Class to define the structure, it would make it more compact, less susceptible to typos, and easier to read. eg
```
class ConversionField:
    def __init__(self, field_name_flat: str, expression_rule):
        self.field_name_flat = field_name_flat
        self.expression_rule = expression_rule

self.conversion_layout = [
    ConversionField("NHS_NUMBER", self.extractor.extract_nhs_number),
    ConversionField("PERSON_FORENAME", self.extractor.extract_person_forename),
    ConversionField("PERSON_SURNAME", self.extractor.extract_person_surname),
    ...
```

delta_backend/src/conversion_layout.py
```diff
@@ -0,0 +1,146 @@
+from extractor import Extractor
+
+class ConversionLayout:
+    def __init__(self, extractor: Extractor):        
+        self.extractor = extractor
+        self.conversion_layout = [
+          {
+              "fieldNameFlat": "NHS_NUMBER",
```
nhsdevws, line 8: Field names should be defined as enum / class-with-constants rather than Magic Strings

nhsdevws (commented):
Field management - use enums / constants is pretty much a `should`, magic strings can cause problems

delta_backend/src/conversion_layout.py
robertnovac1: That's a very good suggestion

delta_backend/src/conversion_layout.py
robertnovac1: Created enums

delta_backend/src/conversion_layout.py
robertnovac1: Done

delta_backend/src/extractor.py
```diff
@@ -0,0 +1,409 @@
...
+    
+    def _convert_date(self, field_name, date, format) -> str:
+        """
+        Convert a date string according to match YYYYMMDD format.
+        """
+        if not date or not isinstance(date, str):
+            self._log_error(field_name, date, "Invalid value. Must be non empty string")
+            return ""
+        try:
+            dt = datetime.fromisoformat(date)
```
Akol125, line 182: there might be a slight problem for the recorded date field - I just caught it while trying to see how API passes data to the Delta. 
date strings with YYYY-MM-DD:Thh:mm:ss:zz not behaving as expected coming out as YYYYMMDD rather than YYYYMMDDThhmmsszz. So bascially recorded is meant to accept both formats (YYYYMMDD & YYYYMMDDThhmmsszz) , it only defers from occurencedatetime because it wouldn't append zeros when a YYYYMMDD data goes in. This might help - https://nhsd-jira.digital.nhs.uk/browse/VED-227.

robertnovac1: That can be included in another PR. This is becoming too large and blocking other changes

delta_backend/src/extractor.py
```diff
@@ -0,0 +1,409 @@
...
+    def __init__(self, fhir_json_data, report_unexpected_exception = True):
+        self.fhir_json_data = json.loads(fhir_json_data) if isinstance(fhir_json_data, str) else fhir_json_data
+        self.report_unexpected_exception = report_unexpected_exception
+        self.error_records = []
+        
+    def _get_patient(self):
+        contained = self.fhir_json_data.get("contained", [])
+        return next((c for c in contained if isinstance(c, dict) and c.get("resourceType") == "Patient"), "")
+
+    def _get_valid_names(self, names, occurrence_time):
```
mfjarvis, line 30: The spec has one extra rule - only consider names which have a `given` and `family`. I feel like we should question this though - follow up ticket?

delta_backend/src/extractor.py
```diff
@@ -0,0 +1,409 @@
...
+    
+    def _get_practitioner_names(self):
+        contained = self.fhir_json_data.get("contained", [])
+        occurrence_time = self._get_occurance_date_time()
+        practitioner = next((c for c in contained if isinstance(c, dict) and c.get("resourceType") == "Practitioner"), None)
+        if not practitioner or "name" not in practitioner:
+            return "", ""
+
+        practitioner_names = practitioner.get("name", [])
+        valid_practitioner_names = [n for n in practitioner_names if "given" in n or "family" in n]
```
mfjarvis, line 60: The spec says we should have a `given` AND `family`. I feel like we should question this though - follow up ticket?

delta_backend/src/extractor.py
```diff
@@ -0,0 +1,409 @@
...
+                    return value_string
+
+        return coding.get("display", "")
+    
+    def _get_site_information(self):
+        performers = self.fhir_json_data.get("performer", [])
+        if not isinstance(performers, list) or not performers:
+            return "", ""
+
+        valid_performers = [p for p in performers if "actor" in p and "identifier" in p["actor"]]
```
mfjarvis, line 131: The spec says we should check for `system` as well

mfjarvis: Ah maybe not

mfjarvis, line 131: The spec says we should prefer performers with a `system` unless there's only one specified - follow up ticket?

delta_backend/src/extractor.py
```diff
@@ -0,0 +1,409 @@
...
+        occurrence_time = self._get_occurance_date_time()
+        patient = self._get_patient()
+        
+        addresses = patient.get("address", [])
+        if not isinstance(addresses, list) or not addresses:
+            return self.DEFAULT_POSTCODE
+
+        valid_addresses = [a for a in addresses if "postalCode" in a and self._is_current_period(a, occurrence_time)]
+        if not valid_addresses:
+            return self.DEFAULT_POSTCODE
```
mfjarvis, line 254: If there's only one address, the spec says to use it even if it's not current at the time of vaccination - follow up ticket?

delta_backend/src/extractor.py
```diff
@@ -0,0 +1,409 @@
...
+        try:
+            occurrence_time = datetime.fromisoformat(self.fhir_json_data.get("occurrenceDateTime", ""))
+            if occurrence_time and occurrence_time.tzinfo is None:
+                occurrence_time = occurrence_time.replace(tzinfo=timezone.utc)
+                return occurrence_time
+            return occurrence_time
+        
+        except Exception as e:
+            message = "DateTime conversion error [%s]: %s" % (e.__class__.__name__, e)
+            error = self._log_error("DATE_AND_TIME", message, e, code=exception_messages.UNEXPECTED_EXCEPTION)
```
mfjarvis, line 96: Could use field name constants here and a few other places in this file

delta_backend/src/extractor.py
```diff
@@ -0,0 +1,409 @@
...
+    
+    def extract_site_code(self):
+        return self._get_site_information()[0]
+    
+    def extract_site_code_type_uri(self):
+        return self._get_site_information()[1]
+        
+    def extract_unique_id(self):
+        identifier = self.fhir_json_data.get("identifier", [])
+        if identifier and len(identifier) == 1:
```
mfjarvis, line 279: Not sure we need the length check?

delta_backend/src/extractor.py
```diff
@@ -0,0 +1,409 @@
...
+    def extract_recorded_date(self) -> str:
+        date = self.fhir_json_data.get("recorded", "")
+        return self._convert_date("RECORDED_DATE", date, self.DATE_CONVERT_FORMAT)
+    
+    def extract_primary_source(self) -> bool | str: 
+        primary_source = self.fhir_json_data.get("primarySource")
+        
+        if isinstance(primary_source, bool):
+            return primary_source
+        if str(primary_source).strip().lower() == "true":
```
mfjarvis, line 304: Don't think we should ever get something other than a boolean

mfjarvis (approved):
Looks really good. Couple of minor differences from the spec, but happy for these to be looked at later

delta_backend/src/extractor.py
robertnovac1: Correct - it's always boolean in backend api

delta_backend/src/extractor.py
robertnovac1: I think this is being handled below

PR #526, opened 2025-05-27
VED-306 Delta refactoring follow up tasks

robertnovac1:
## Summary
* Routine Change

- Primary source accepting only boolean
- Unit test refactoring 
- Fixed: If there's only one address, the spec says to use it even if it's not current at the time of vaccination
- Fixed: Change extraction logic for the practitioner and patient name to extract instance that have both given and family names. Current logic checks if either given or family name exist
- Used constants for flat field names

## Reviews Required
* [x] Dev

## Review Checklist

### DIFF
--- a/backend/src/models/utils/pre_validator_utils.py
+++ b/backend/src/models/utils/pre_validator_utils.py
@@ -113,16 +113,16 @@ class PreValidation:
         if not isinstance(field_value, str):
             raise TypeError(f"{field_location} must be a string")
 
-        error_message = f"""{field_location} must be a valid datetime in one of the following formats:
-        - 'YYYY-MM-DD' — Full date only
-        - 'YYYY-MM-DDThh:mm:ss' — Full date and time without milliseconds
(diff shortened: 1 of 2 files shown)


nhsdevws (dismissed):

nhsdevws (dismissed):

mfjarvis (dismissed):

delta_backend/src/extractor.py
```diff
@@ -48,8 +48,11 @@ def _get_person_names(self):
         person_forename = " ".join(selected_name.get("given", []))
         person_surname = selected_name.get("family", "")
 
-        return person_forename, person_surname
-
+        if person_forename and person_surname:
```
mfjarvis, line 51: What if there were two valid names at the occurrence datetime, the first only had a `given` but the second had a `given` and `family`?

mfjarvis: Do we need something like this instead?
```
        complete_names = [n for n in names if n.get("given", []) and n.get("family", "")]
        selected_name = self._get_valid_names(complete_names, occurrence_time)
```

Akol125 (approved):

PR #575, opened 2025-06-10
VED-294: Fix potential duplication in the delta.

mfjarvis:
## Summary
* Routine Change

From the AWS documentation:

> A Lambda consumer for a DynamoDB stream doesn't guarantee exactly once delivery and may lead to occasional duplicates. Make sure your Lambda function code is idempotent to prevent unexpected issues from arising because of duplicate processing.

Changes:
- Use the `eventID` from the stream event as the primary key of the delta record instead of creating a new unique ID for each record.
- Add a condition to the put so we only insert the delta record if it does not already exist. 
- Treat exceptions due to duplicate records as success.
- Increase timeout to handle processing a full batch of 100 records.
- Split main handler code into multiple functions for readability.

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist

### DIFF
--- a/recordprocessor/src/file_level_validation.py
+++ b/recordprocessor/src/file_level_validation.py
@@ -65,7 +65,7 @@ def move_file(bucket_name: str, source_file_key: str, destination_file_key: str)
 
 
 @file_level_validation_logging_decorator
-def file_level_validation(incoming_message_body: dict) -> None:
+def file_level_validation(incoming_message_body: dict) -> dict:
     """
     Validates that the csv headers are correct and that the supplier has permission to perform at least one of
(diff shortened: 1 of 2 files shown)


nhsdevws (dismissed):
Good Tidy up. Couple of points for future 
- Unit tests need on-going attention with dependency injection, particularly on the handler
- Maybe keep the term `handle` for the aws handler. Calls from process_record retain a similar convention, ie 
- - handle_remove - process_remove
- - handle_skipped - process_skip
- - handle_create_update_delete - process_change

Small point.

PR #722, opened 2025-08-07
Ved 699 sonar coverage all projects

robertnovac1:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist


redis_sync/pyproject.toml
```diff
@@ -17,7 +17,7 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.11"
+python = "~3.11"
```
dlzhry2, line 20: The `poetry.lock` file also needs to be updated for the `redis_sync` project. Can see that the unit tests for this one fail to run.

.github/workflows/sonarcloud.yml
```diff
@@ -21,13 +21,9 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Install poetry
-        run: pip install poetry==2.1.2
-
       - uses: actions/setup-python@v5
         with:
           python-version: 3.11
-          cache: 'poetry'
```
dlzhry2, line 30: This fixes the issue, but one thing we lose with this change is the caching of poetry package installations. Maybe we could use the Github cache action?

Or maybe we just go ahead with this as a temporary fix and when the setup-python action is fixed then we revert back to using it. (We probably would need a ticket to remember to do so).

By the way, I had realised and shared on James' ticket that the root cause is actually an issue with the `setup-python` Github action - I linked to it here - https://github.com/NHSDigital/immunisation-fhir-api/pull/720#pullrequestreview-3097877419

This kicked in about 4/5 days ago.

dlzhry2 (approved):
Looks good. Just need the poetry lock in the `redis_sync` project. And potentially needs a ticket to remove this temporary fix once the team that manages the Github Action for python-setup has fixed it.

PR #732, opened 2025-08-12
VED-59 Provide millisecond precision for ack processor logs

dlzhry2nhs:
## Summary
* Routine Change

As discussed on ticket - see link in comments - the logs created by the convert_message_to_ack_row_logging_decorator currently are measure in seconds and rounded to 5 d.p.

As the logs are generated and the time is measured for each individual ack record the time is very short indeed and sometimes results into the format seen in the ticket:

image

Therefore, discussed and agreed with Martin that for these specific logs we will set them to millisecond precision given the nature of what we are timing.

See the ticket itself for the initial dev testing comments. You should now be able to see that the time is presented correctly.

## Reviews Required
* [ ] Dev

## Review Checklist


ack_backend/src/logging_decorators.py
```diff
@@ -22,22 +22,29 @@ def send_log_to_firehose(log_data: dict) -> None:
...
+    base_log_data: dict,
+    additional_log_data: dict,
+    use_ms_precision: bool = False,
+    is_error_log: bool = False
 ) -> None:
     """Generates log data which includes the base_log_data, additional_log_data, and time taken (calculated using the
     current time and given start_time) and sends them to Cloudwatch and Firehose."""
-    log_data = {**base_log_data, "time_taken": f"{round(time.time() - start_time, 5)}s", **additional_log_data}
+    seconds_elapsed = time.time() - start_time
+    formatted_time_elapsed = f"{round(seconds_elapsed * 1000, 5)}ms" if use_ms_precision else \
```
robertnovac1, line 34: Is there a requirement to not use ms precision? Thinking it might be worth always using ms_precision and then you won't need the if/else.

dlzhry2nhs: Yes, from what I understood from Martin. For example, the other logging decorator in this file wraps the actual lambda handler. This can take more than one second so seemed okay. From what he explained, ITOC's Splunk queries are flexible so it does not matter if some are in seconds and some in milliseconds.

Don't see why we couldn't look into making everything consistent though in future. I was just constraining the change to this specific area as the time taken is very small in this section of code.

robertnovac1 (approved):

PR #748, opened 2025-08-18
VED-124 Batch Splunk

JamesW1-NHS:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist

### DIFF
--- a/recordprocessor/tests/test_logging_decorator.py
+++ b/recordprocessor/tests/test_logging_decorator.py
@@ -21,6 +21,7 @@ with patch.dict("os.environ", MOCK_ENVIRONMENT_DICT):
     from errors import InvalidHeaders, NoOperationPermissions
     from logging_decorator import send_log_to_firehose, generate_and_send_logs
     from file_level_validation import file_level_validation
+    from batch_processor import ingestion_progress
 
 
 from tests.utils_for_recordprocessor_tests.utils_for_recordprocessor_tests import GenericSetUp, GenericTearDown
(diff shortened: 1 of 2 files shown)


recordprocessor/src/batch_processor.py
```diff
@@ -8,9 +8,15 @@
 from send_to_kinesis import send_to_kinesis
 from clients import logger
 from file_level_validation import file_level_validation
+from logging_decorator import ingestion_logging_decorator
 from errors import NoOperationPermissions, InvalidHeaders
 
 
+@ingestion_logging_decorator
+def ingestion_progress(finished: bool, message_body: dict, row_count: int = 0, start_time: float = 0.0) -> str:
```
dlzhry2nhs, line 16: I think it might be unnecessary to create this function. Almost all the arguments are unused by the function, so we are only really creating it for the purpose of the decorator. While the codebase technically also uses this pattern elsewhere e.g. the logging decorators often use the function's result, it would be nice to avoid introducing it in new places if we can.

The reason I'm not so keen, is that it breaks encapsulation - the function is no longer just thinking about what _it_ needs to do but also what _external_ things like decorators might require.

recordprocessor/src/batch_processor.py
```diff
@@ -8,9 +8,15 @@
...
+    return("Ingestion finished" if finished else "Ingestion started")
+
+
 def process_csv_to_fhir(incoming_message_body: dict) -> None:
```
dlzhry2nhs, line 20: I think we could just wrap the `process_csv_to_fhir` function, taking a reading for the time at the start and time at the end. This will give us the total time to convert all the records to FHIR and send to Kinesis, which I think is one of the acceptance criteria of the ticket.

We could have it return data that the wrapper could use e.g. number of rows, filename/message metadata etc.

The other option would be to simply have a helper function that logs in-line but we would need to ensure that it would send to Splunk as well as just Cloudwatch.

recordprocessor/src/batch_processor.py
```diff
@@ -55,6 +65,12 @@ def process_csv_to_fhir(incoming_message_body: dict) -> None:
...
         logger.info("Total rows processed: %s", row_count)
 
+    logger.info(ingestion_progress(
+                    finished=True,
+                    message_body=interim_message_body,
+                    row_count=row_count,
+                    start_time=ingestion_start_time))
+
 
 def main(event: str) -> None:
```
dlzhry2nhs, line 75: We could probably move the timers and logging from main if we introduce the logging decorator.

dlzhry2nhs (commented):
Looks good so far. Couple of suggestions for minor improvements.

Comment on the wider piece of work: this PR is currently only adding logs to measure the time taken for the batch ECS processor to convert records to FHIR and send for onward processing. However, I am not sure if we would be able to determine the full e2e processing time from these logs alone. See https://nhsd-confluence.digital.nhs.uk/spaces/Vacc/pages/1035417078/New+Batch+Ingestion+Design#NewBatchIngestionDesign-NewBatchIngestionDesign for more context.

Line 88 in ack_backend/src/update_ack_file.py is when a file is fully processed i.e. the batch changes have been made and saved to the database and an acknowledgement has been created for the supplier. Martin may be the person to clarify the requirements with, but I cannot see unless we have a log here that also include the filename and message ID (so you can correlate and diff the times in Splunk) how you could know the full end-to-end time taken.

PR #767, opened 2025-08-26
VED-737-Search-API-validate parameters

nhsdevws:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist


backend/src/parameter_parser.py
```diff
@@ -162,3 +165,53 @@ def create_query_string(search_params: SearchParams) -> str:
...
+def check_route_parameters(query_params, body, valid_params: list):
+    try:
+
+        query_params = query_params or {}
+        body_dict = body_to_dict(body)
+        # merge query and body parameters
+        all_params = {**query_params, **body_dict}
+
+        found = False
+        for param in all_params:
```
dlzhry2nhs, line 187: We could avoid 2 for loops if we set `found` when we find a valid param in the for loop below.

backend/src/search_imms_handler.py
```diff
@@ -24,34 +27,20 @@ def search_imms_handler(event: events.APIGatewayProxyEventV1, _context: context_
...
-                or body_has_immunization_element
-            ):
+        has_body = body is not None
+        has_query_params = query_params is not None and query_params != {}
+        if has_query_params or has_body:
+            parsed_body = get_parsed_body(body)
+            if is_immunization_by_identifier(query_params, parsed_body):
                 return controller.get_immunization_by_identifier(event)
+            elif is_search_immunizations(query_params, parsed_body):
+                return controller.search_immunizations(event)
```
dlzhry2nhs, line 40: Sorry, I am not that familiar with this area of the code yet, but this looks to me like a potentially functional change.

Previously, this whole block was basically checking for a situation in which you have the required params for `get_immunization_by_identifier` and if so, calling it. Else, it would go straight to `response = controller.search_immunizations(event)` and then execute the remaining code.

By adding this `elif` condition it's a slight change, but maybe that's okay. I would have the following questions:

- Under what circumstances will line 44 now be called?
- By simply returning the result here at line 40, will we be missing some of the logic from line 46 onwards i.e. checking if the size of the response is over a certain threshold?

dlzhry2nhs (commented):
Couple of comments added. Nice changes and cleans up the body and query string parsing we're doing.

As you'll have seen, the changes have introduced some 400s where 200s were previously expected in the e2e pipeline, so that will need some looking into.

dlzhry2nhs:
PR still needed? If not please close.

PR #770, opened 2025-08-26
VED-714 rearchitect to prevent batch race conditions

dlzhry2nhs:
## Summary
* Routine Change - improves batch processing reliability

See ticket and especially the architecture redesign page for the full context: https://nhsd-confluence.digital.nhs.uk/spaces/Vacc/pages/1161762503/Immunisation+FHIR+API+-+Batch+Ingestion+Improvements

In short, it resolves race conditions in the filename processor around both checking for duplicate files and ensuring that only one event is in progress for a given supplier + vacc type. However, the major benefit of it is to simplify the design, improve reliability and remove the unusual pattern whereby Lambdas were directly invoking each other.

**Infra changes:**
- added new batch processor filter lambda function with required permissions
- added new batch file created SQS fifo queue
- removed permissions no longer required by several other Lambda functions that have been refactored

**Code changes:**
- Refactored the `ack_backend` Lambda so it no longer uses the backdoor to invoke the filename proc Lambda
- Refactored the `recordprocessor` ECS task so it marks an Audit Entry as Preprocessed once all files have been sent to Kinesis, and to also no longer use the backdoor to directly invoke the filename proc Lambda.
- Refactored the `filenameprocessor` Lambda so it no longer makes a backdoor available for direct invocations, no longer performs duplicate file checks (due to potential race conditions) and no longer checks if a file is already processing (again, due to race conditions as the events were coming directly from the S3 create event).

Added a new Lambda function to handle batch process filtering, as per the Confluence diagram. It will:
- consume from the new FIFO queue. This eliminates the race condition, as for a given message group (supplier + vaccination type), there will only be one event processed at a time.
- check the audit table for any duplicate files
- check the audit table for any events currently processing for the given message group
- if all clear, it will send the event to the next queue for processing by ECS and mark the audit status accordingly

## Reviews Required
* [ ] Dev

## Review Checklist


terraform/sqs_batch_file_created.tf
```diff
@@ -0,0 +1,7 @@
+# FIFO SQS Queue - targeted by Filename Processor Lambda function
+resource "aws_sqs_queue" "batch_file_created" {
+  name                        = "${local.short_prefix}-batch-file-created-queue.fifo"
+  fifo_queue                  = true
+  content_based_deduplication = true # Optional, helps with deduplication
+  visibility_timeout_seconds  = 60 # TODO - discuss and refine this + DDB TTL
```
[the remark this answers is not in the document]
dlzhry2nhs: See Lambda Concurrency behaviour with FIFO here: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/fifo-queue-lambda-behavior.html
It scales out on a per message group basis.

recordprocessor/src/constants.py
```diff
@@ -53,7 +53,8 @@ class FileStatus:
 
     QUEUED = "Queued"
     PROCESSING = "Processing"
-    PROCESSED = "Processed"
+    PREPROCESSED = "Preprocessed"  # All entries in file converted to FHIR and forwarded to Kinesis
```
dlzhry2nhs, line 56: Thought it useful to introduce an additional status. We needed to mark the event as something other than 'Processing' so that the next event can be picked up.

However, 'Processed' might be slightly misleading as the changes have not yet been written to the DB. The ECS record processor merely converts the records to FHIR and then sends to Kinesis Please shout if any objections to this.

batch_processor_filter/src/lambda_handler.py
```diff
@@ -0,0 +1,23 @@
+import json
+from aws_lambda_typing import events, context
+
+from batch_file_created_event import BatchFileCreatedEvent
+from batch_processor_filter_service import BatchProcessorFilterService
+from exceptions import InvalidBatchSizeError
+
+
+service = BatchProcessorFilterService()
```
dlzhry2nhs, line 9: At some point, maybe beyond the pipeline refactoring and some upcoming releases, think we should move our other Lambdas (especially backend API) to make use of instantiation outside of the handler so it can be reused within an execution environment. Especially useful for things like Table/DB resources: https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html see point 1

Note: can only be used for objects that are threadsafe i.e. do not modify shared state.

batch_processor_filter/src/batch_audit_repository.py
```diff
@@ -0,0 +1,46 @@
+import boto3
+from boto3.dynamodb.conditions import Key
+
+from constants import AUDIT_TABLE_NAME, REGION_NAME, AUDIT_TABLE_FILENAME_GSI, AuditTableKeys, FileStatus, \
+    AUDIT_TABLE_QUEUE_NAME_GSI
+
+
+class BatchAuditRepository:
+    """Batch audit repository class."""
+    _DUPLICATE_CHECK_FILE_STATUS_CONDITION = (
```
JamesW1-NHS, line 10: Is this a good name, given that the file status can also (technically) be QUEUED?

batch_processor_filter/src/batch_file_repository.py
```diff
@@ -0,0 +1,60 @@
...
+    _SOURCE_BUCKET_NAME: str = SOURCE_BUCKET_NAME
+    _ACK_BUCKET_NAME: str = ACK_BUCKET_NAME
+
+    def __init__(self):
+        self._s3_client = boto3.client('s3')
+
+    @staticmethod
+    def _create_ack_failure_data(batch_file_created_event: BatchFileCreatedEvent) -> dict:
+        return {
+            "MESSAGE_HEADER_ID": batch_file_created_event["message_id"],
```
JamesW1-NHS, line 23: should we implement this dict as a constant somewhere, and just fill its variable values when we call the method?

batch_processor_filter/src/constants.py
```diff
@@ -0,0 +1,31 @@
+import os
+from enum import StrEnum
+
+REGION_NAME = "eu-west-2"
```
JamesW1-NHS, line 4: Strictly we should give os.getenv the opportunity to override the region, e.g. _os.getenv("AWS_REGION", "eu-west-2")_
but I admit it's pedantic in this case

terraform/batch_processor_filter_lambda.tf
```diff
@@ -0,0 +1,303 @@
...
+        "Action" : [
+          "ecr:BatchGetImage",
+          "ecr:DeleteRepositoryPolicy",
+          "ecr:GetDownloadUrlForLayer",
+          "ecr:GetRepositoryPolicy",
+          "ecr:SetRepositoryPolicy"
+        ],
+        "Condition" : {
+          "StringLike" : {
+            "aws:sourceArn" : "arn:aws:lambda:eu-west-2:${var.immunisation_account_id}:function:${local.short_prefix}-batch-processor-filter-lambda"
```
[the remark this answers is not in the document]
dlzhry2nhs: Good suggestion - actioned.

batch_processor_filter/src/batch_audit_repository.py
dlzhry2nhs: Yes, this is the intended functionality i.e. the statuses in which we flag a file as duplicate (assuming the name is matching - see usage at line 22 in the same file).

Reasoning - let's say 3 files are created at a similar time by mistake, all with the same name. At first all will be in QUEUED state. We will want to process the first one. However, any subsequent one should be marked as a duplicate.

Therefore the logic is, if there exists a file with the same name which has already been processed, preprocessed or is currently processing, then mark it as a duplicate.

batch_processor_filter/src/batch_file_repository.py
dlzhry2nhs: I'm not sure what value this would add. I don't like Python dictionary mutability, and writing to an existing constant seems like it has potential for bad side effects.

Although it is maybe a little clunky (this is moved from where it used to be in the filename processor, albeit with a little tidying), it ensures there are no mutability concerns. You get a fresh dict each time with dynamic values.

PR #777, opened 2025-08-28
VED-736: refactor response for search identifier

Akol125:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist


backend/tests/test_fhir_service.py
```diff
@@ -202,11 +202,11 @@ def test_get_immunization_by_id(self):
         self.imms_repo.get_immunization_by_id.return_value = {"Resource": create_covid_19_immunization(imms_id).dict()}
 
         # When
-        service_resp = self.fhir_service.get_immunization_by_id(imms_id, "COVID19:read")
+        service_resp = self.fhir_service.get_immunization_by_id(imms_id, "COVID19.CRUDS")
```
dlzhry2nhs, line 205: I'm not sure I follow, but I might be missing some context.
In the code, we are changing the format of a response. However, in the tests we are just changing the format of some permissions.

Why are we making these changes to the tests? And where is the test that proves the new changes to the response body?

Akol125: its just something I saw in the test that was wrong and should have been changed. we have the old format permissions, which has been changed to the new permission format

dlzhry2nhs: That's great, thanks for the context. Makes sense re the permissions change then :)

Is there a test to assert the new response format?

Akol125: yes we have used a module Filter.search which already asserts the response format, while the generic utils "form_json" used to wrap the response already asserts that each parent dictionary response are in the right format and the search service asserts if its a bundle.

dlzhry2nhs: Yes, my point is that we have changed the format of what `get_immunization_by_identifier` returns yet we have not actually changed any assertions either in `e2e` tests or the unit tests.

It's just a bit unusual. Have you manually checked the output from an API request? I would have thought this would start failing in e2e but maybe I'm missing something: `e2e/utils/resource.py`

PR #784, opened 2025-09-01
VED-189 Refactor authorisation

dlzhry2nhs:
## Summary
* Routine Change

Started off as a review of the state of authorisation in line with the proposed changes in Confluence: https://nhsd-confluence.digital.nhs.uk/spaces/Vacc/pages/1035417415/Immunisation+FHIR+API+-+Move+System+Authorisation+Checks+from+Apigee

Much of the work had been done to move supplier checks to Redis for the backend API. The only thing left to do was remove checks for obsolete headers and tidy/consolidate authorisation into a single layer rather than bits here and there.

**Key Changes:**

**Proxy:**
- Only forward the supplier system custom attribute. Others such as `VaccineTypePermissions` or `AuthenticationType` are no longer forwarded and can be safely removed from Apigee apps.

**Backend code:**
- Removed the Controller layer authorization which checked that the `AuthenticationType` header was present - no longer needed.
- Removed permission checking from the Repository layer. Authorisation checks should be at the Controller or Service level (I have gone with service). Moved that logic into the new authorisation class.
- Added authorisation class (and injected into the service layer) to retrieve supplier permissions and add checks against requested operations.
- Tests updated accordingly. Plus added tests for the new authoriser and where it is used in the fhir_service.

Ultimately, this is purely refactoring so no behaviour should change. However, I have tidied code where I can but tried not to make this PR too large hence why I raised https://nhsd-jira.digital.nhs.uk/browse/VED-747 for the update journey which really needs some refactoring.

Some examples of changes I made as I went were to move the check for existing in the Delete journey to the service layer. The controllers, in the main, are doing too much business logic above and beyond simple parsing, validating and marshalling of input specific (AWS event) format. That business logic belongs in the service layer.

## Reviews Required
* [ ] Dev

## Review Checklist

### DIFF
--- a/backend/src/fhir_service.py
+++ b/backend/src/fhir_service.py
@@ -151,7 +151,8 @@ class FhirService:
 
         vaccination_type = get_vaccine_type(immunization)
 
-        # If the user is updating the vaccination_type, they must have permissions for both the new and old type
+        # If the user is updating the resource vaccination_type, they must have permissions for both the existing and
+        # new type. In most cases it will be the same, but it is possible for users to update the vacc type
         if not self.authoriser.authorise(supplier_system, ApiOperationCode.UPDATE,
                                          {vaccination_type, existing_resource_vacc_type}):
             raise UnauthorizedVaxError()


backend/src/fhir_service.py
```diff
@@ -121,34 +122,45 @@ def create_immunization(
...
         immunization["id"] = imms_id
 
         patient = self._validate_patient(immunization)
         if "diagnostics" in patient:
-            return (None, patient, None)
+            return None, patient, None
+
+        vaccination_type = get_vaccine_type(immunization)
+
+        # If the user is updating the vaccination_type, they must have permissions for both the new and old type
```
Akol125, line 154: sorry Daniel I don't quite understand this part, do you mean old and new authorization or vaccine types????

Akol125: Yeah i would definitely let you know when I am done, thanks for your patience, also trying to resolve an issue on pydantic fhir models from my ticket hence the delay.. should be through shortly

Akol125 (approved):

backend/tests/test_fhir_controller.py
```diff
@@ -1221,15 +1041,13 @@ def test_update_deletedat_immunization_with_version(self, mock_get_supplier_perm
...
         self.service.reinstate_immunization.assert_called_once_with(
-            imms_id, json.loads(imms), 1, ["COVID19.CRUDS"], "Test"
+            imms_id, json.loads(imms), 1, "COVID19", "Test"
         )
         self.assertEqual(response["statusCode"], 200)
         self.assertEqual(response["headers"]["E-Tag"], 2)
 
-    @patch("fhir_controller.get_supplier_permissions")
-    def test_update_deletedat_immunization_without_version(self, mock_get_supplier_permissions):
+    def test_update_deletedat_immunization_without_version(self):
```
JamesW1-NHS, line 1049: Pedantic point: according to naming conventions, given that the list key is 'DeletedAt', strictly the function name should be 'test_updated_deleted_at_immunization...'
Also applies to e.g. 'test_validate_imms_id_invalid_vaccinetype' (should strictly be '..._vaccine_type', and there may be others

JamesW1-NHS (approved):

PR #816, opened 2025-09-12
VED-755: Handling NHS Number confusions

Akol125:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist

### DIFF
--- a/backend/src/log_structure.py
+++ b/backend/src/log_structure.py
@@ -42,7 +42,6 @@ def function_info(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
         event = args[0] if args else {}
-        logger.info(f"Event: {event}")
         headers = event.get("headers", {})
         correlation_id = headers.get("X-Correlation-ID", "X-Correlation-ID not passed")
         request_id = headers.get("X-Request-ID", "X-Request-ID not passed")
@@ -64,7 +63,6 @@ def function_info(func):
         start = time.time()
         try:
             result = func(*args, **kwargs)
-            logger.info(f"Result:{result}")
             end = time.time()
             log_data["time_taken"] = f"{round(end - start, 5)}s"
             log_data.update(_log_data_from_body(event))
--- a/backend/tests/test_log_structure_wrapper.py
+++ b/backend/tests/test_log_structure_wrapper.py
@@ -23,6 +23,14 @@ class TestFunctionInfoWrapper(unittest.TestCase):
     def mock_function_raises(_event, _context):
         raise ValueError("Test error")
 
+    def extract_all_call_args_for_logger(self, mock_logger) -> list:
+        """Extracts all arguments for logger.*."""
+        return (
+            [ args[0] for args, _ in mock_logger.info.call_args_list ]
+            + [ args[0] for args, _ in mock_logger.warning.call_args_list ]
+            + [ args[0] for args, _ in mock_logger.error.call_args_list ]
+        )
+
     def test_successful_execution(self, mock_logger, mock_firehose_logger):
         # Arrange
         test_correlation = "test_correlation"
@@ -64,6 +72,37 @@ class TestFunctionInfoWrapper(unittest.TestCase):
         self.assertEqual(logged_info['resource_path'], test_resource_path)
         self.assertEqual(logged_info['local_id'], '12345^http://test')
         self.assertEqual(logged_info['vaccine_type'], 'FLU')
+
+    def test_successful_execution_pii(self, mock_logger, mock_firehose_logger):
+        """Pass personally identifiable information in an event, and ensure that it is not logged anywhere."""
+        # Arrange
+        test_correlation = "test_correlation"
+        test_request = "test_request"
+        test_supplier = "test_supplier"
+        test_actual_path = "/test"
+        test_resource_path = "/test"
+
+        self.mock_redis_client.hget.return_value = "FLU"
+        wrapped_function = function_info(self.mock_success_function)
+        event = {
+            'headers': {
+                'X-Correlation-ID': test_correlation,
+                'X-Request-ID': test_request,
+                'SupplierSystem': test_supplier
+            },
+            'path': test_actual_path,
+            'requestContext': {'resourcePath': test_resource_path},
+            'body': "{\"identifier\": [{\"system\": \"http://test\", \"value\": \"12345\"}], \"contained\": [{\"resourceType\": \"Patient\", \"id\": \"Pat1\", \"identifier\": [{\"system\": \"https://fhir.nhs.uk/Id/nhs-number\", \"value\": \"9693632109\"}]}], \"protocolApplied\": [{\"targetDisease\": [{\"coding\": [{\"system\": \"http://snomed.info/sct\", \"code\": \"840539006\", \"display\": \"Disease caused by severe acute respiratory syndrome coronavirus 2\"}]}]}]}"
+        }
+
+        # Act
+        result = wrapped_function(event, {})
+
+        # Assert
+        self.assertEqual(result, "Success")
+
+        for logger_info in self.extract_all_call_args_for_logger(mock_logger):
+            self.assertNotIn("9693632109", str(logger_info))
 
     def test_exception_handling(self, mock_logger, mock_firehose_logger):
         # Arrange
--- a/config/prod/permissions_config.json
+++ b/config/prod/permissions_config.json
@@ -21,7 +21,10 @@
   },
   {
     "supplier": "RAVS",
-    "permissions": ["RSV.RS"],
+    "permissions": [
+      "MMR.CRUDS",
+      "RSV.RS"
+    ],
     "ods_codes": ["X26", "X8E5B"]
   },
   {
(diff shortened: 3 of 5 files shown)


lambdas/id_sync/src/record_processor.py
```diff
@@ -1,61 +1,172 @@
...
-    else:
-        logger.info("No NHS number found in event record")
-        return {"status": "error", "message": "No NHS number found in event record"}
 
+    logger.info("No NHS number found in event record")
+    return {"status": "error", "message": "No NHS number found in event record"}
 
-def process_nhs_number(nhs_number: str) -> Optional[str]:
+
+def process_nhs_number(nhs_number: str) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
dlzhry2nhs: It is good that you identified this. Ideally we should try to break down our functions/methods sufficiently as not to trigger this, although exceptions can be made in specific circumstances. Hopefully we can resolve this one here.

lambdas/id_sync/src/record_processor.py
```diff
@@ -1,61 +1,176 @@
...
+
+        if not match_found:
+            logger.info("process_nhs_number: No IEDS items matched PDS demographics. Skipping update for %s", nhs_number)
+            response = {
+                "status": "success",
+                "message": "No IEDS items matched PDS demographics; update skipped",
+            }
         else:
-            response = {"status": "success", "message": "No update required"}
+            response = ieds_update_patient_id(nhs_number, new_nhs_number)
```
dlzhry2nhs, line 100: When we update, do we just need to change the PK or do we also amend the FHIR JSON or do we leave it alone?

JamesW1-NHS (dismissed):

lambdas/id_sync/src/ieds_db_operations.py
```diff
@@ -141,3 +141,19 @@ def get_items_from_patient_id(id: str, limit=BATCH_SIZE) -> list:
             nhs_numbers=[patient_pk],
             exception=e
         )
+
```
dlzhry2nhs, line 144: https://screening-discovery.slack.com/archives/C08AVB8MEBG/p1758033638402969 relates to the above functions: `get_items_from_patient_id` and `ieds_update_patient_id` but from the code we have read, we believe:

A: the update request will not actually work.
B: if a patient has more than 25 vaccination records then only up to 25 will be updated. The others will be ignored.

As part of this ticket's dev testing, could you have a look into this. If you think you can handle it as part of this ticket (assuming it does not work) then feel free to pick it up. Otherwise we can raise a ticket to track it separately.

lambdas/id_sync/src/record_processor.py
```diff
@@ -1,61 +1,176 @@
...
-        # if patient NHS != id, update patient index of vax events to new number
-        if patient_details_id != nhs_number and patient_details_id:
-            logger.info(f"process_nhs_number. Update patient ID from {nhs_number} to {patient_details_id}")
-            if ieds_check_exist(nhs_number):
-                logger.info("process_nhs_number. IEDS record found, updating patient ID")
-                response = ieds_update_patient_id(nhs_number, patient_details_id)
-            else:
-                logger.info("process_nhs_number. No ieds record found for: %s", nhs_number)
-                response = {"status": "success", "message": f"No records returned for ID: {nhs_number}"}
+    new_nhs_number = pds_get_patient_id(nhs_number)
```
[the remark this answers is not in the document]
Akol125: yes you are right, query using the old nhs number then the the response carries ["identifier"]["value"] which has the new nhs number. Also we could change the variable names, that's totally fine by me.

lambdas/id_sync/src/id_sync.py
```diff
@@ -1,46 +1,51 @@
...
+"""
 
-@logging_decorator(prefix="id_sync", stream_name=STREAM_NAME)
-def handler(event_data, _):
 
+@logging_decorator(prefix="id_sync", stream_name=STREAM_NAME)
+def handler(event_data: Dict[str, Any], _context) -> Dict[str, Any]:
     try:
-        logger.info("id_sync handler invoked")
         event = AwsLambdaEvent(event_data)
```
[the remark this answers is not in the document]
dlzhry2nhs: Discussed on call - optional. Can do if it does not take too much time, but is an existing issue.

lambdas/id_sync/src/record_processor.py
dlzhry2nhs: At the very least, it is clear from original implementation ticket and Confluence that the FHIR JSON does not need to be updated. We just need to update the PK. Resolving comment. I will also put a comment on JIRA for completeness.

lambdas/id_sync/src/record_processor.py
```diff
@@ -1,61 +1,164 @@
...
+
+    if not new_nhs_number:
+        return log_status("No patient ID found for NHS number", nhs_number)
+
+    if new_nhs_number == nhs_number:
+        return log_status("No update required", nhs_number)
+    logger.info("Update patient ID from %s to %s", nhs_number, new_nhs_number)
+
+    try:
+        pds_details, ieds_details = fetch_demographic_details(nhs_number)
```
dlzhry2nhs, line 53: I wonder if we need to separate this call into one for each. Or we could rename the function to make clear that it retrieves demographic details and vacc details? At the moment it is not very clear as to the intent.

lambdas/id_sync/src/record_processor.py
```diff
@@ -1,61 +1,164 @@
...
+    except Exception as e:
+        logger.exception("process_nhs_number: failed to fetch demographic details: %s", e)
+        return log_status(str(e), nhs_number, "error")
+
+    if not ieds_details:
+        logger.info("No IEDS records returned for NHS number: %s", nhs_number)
```
dlzhry2nhs, line 59: If we are calling them IEDS records, should we also name the variable this ieds_details -> ieds_records. We should make our naming as clear as possible. ieds is an acronym so it might be even better to call it imms_records.

lambdas/id_sync/src/record_processor.py
```diff
@@ -1,61 +1,164 @@
...
+    matching_records, discarded_records = [], []
+    for detail in ieds_details:
+        if demographics_match(pds_details, detail):
+            matching_records.append(detail)
         else:
-            response = {"status": "success", "message": "No update required"}
-    else:
-        response = {"status": "success", "message": f"No patient ID found for NHS number: {nhs_number}"}
-    response.update(base_log_data)
+            discarded_records.append(detail)
```
dlzhry2nhs, line 68: Do we need to store the details in memory if we are discarding it. Maybe we could just have a count of the number discarded if that's all we are looking for.

lambdas/id_sync/src/record_processor.py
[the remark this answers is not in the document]
dlzhry2nhs: Sorry comment applies to a different one. Marking as resolved.

lambdas/id_sync/src/utils.py
```diff
@@ -0,0 +1,16 @@
+from typing import Dict, Any
+
+
+def log_status(msg: str, nhs_number: str | None = None, status: str = "success") -> Dict[str, Any]:
```
dlzhry2nhs, line 4: Function name misleading. It does not perform any logging. The docstring is good though, maybe align the naming to say what it actually does.

lambdas/id_sync/tests/test_ieds_db_operations_conditional.py
```diff
@@ -0,0 +1,76 @@
+import unittest
+from unittest.mock import patch, MagicMock
+
+import ieds_db_operations
+
+
+class TestIedsDbOperationsConditional(unittest.TestCase):
```
dlzhry2nhs, line 7: It's good we have this test. Could we put it in the same file as the existing tests though? It's testing functionality from the same file.

lambdas/id_sync/src/record_processor.py
```diff
@@ -1,61 +1,176 @@
...
+            "status": "success",
+            "message": "No update required",
+            "nhs_number": nhs_number,
+        }
+    logger.info("Update patient ID from %s to %s", nhs_number, new_nhs_number)
+
+    if ieds_check_exist(nhs_number):
+        # Fetch PDS details for demographic comparison
+        try:
+            pds_details = pds_get_patient_details(nhs_number)
```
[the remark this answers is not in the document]
dlzhry2nhs: I have given this some more thought. Maybe it is too much for this ticket, but can we raise a ticket to remove to 2 calls to PDS and replace with 1. It is an external service that already faces a lot of traffic, so I would like to minimise this in future.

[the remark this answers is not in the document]
Akol125: that would be nice, we could have another ticket to tighten up some other checks, including this and add backoffs and retries for the ieds_db operations. A ticket basically to make things more durable and cater for scenarios that might also come up in the test??

lambdas/id_sync/src/record_processor.py
Akol125: resolved to count

lambdas/id_sync/src/record_processor.py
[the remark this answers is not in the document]
dlzhry2nhs: Sounds good.

dlzhry2nhs (approved):

PR #854, opened 2025-09-29
VED-375 Search Parameter Error

Akol125:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist


backend/src/fhir_service.py
```diff
@@ -334,8 +334,6 @@ def search_immunizations(
         """
         # TODO: is disease type a mandatory field? (I assumed it is)
         #  i.e. Should we provide a search option for getting Patient's entire imms history?
-        if not nhs_number_mod11_check(nhs_number):
```
dlzhry2nhs, line 337: Can we do something about the above TODOs? One to check with business or product. Think point 1 we can just remove - from the validation it is clearly a mandatory field.

From point 2, I think this might be beyond the scope of our API but it might be worth checking. Once we know, let's remove it because we should try to get rid of comments and TODOs like this in the codebase unless it actually needs action.

backend/src/parameter_parser.py
```diff
@@ -36,6 +37,90 @@ def __repr__(self):
         return str(self.__dict__)
 
 
+def process_mandatory_params(params: ParamContainer) -> tuple[str, list[str]]:
+    """Validate mandatory params and return (patient_identifier, vaccine_types).
+    Raises ParameterException for any validation error.
+    """
+    # patient.identifier
```
dlzhry2nhs, line 44: This could be abstracted into a standalone function `process_patient_identifier_param`. This will make the function more SOLID and readable.

Have it do all of this validation, raise the ParameterException in the relevant place and return the patient_identifier if it is valid.

backend/src/parameter_parser.py
```diff
@@ -36,6 +37,90 @@ def __repr__(self):
...
+    if len(patient_identifier_parts) != 2 or patient_identifier_parts[0] != patient_identifier_system:
+        raise ParameterException(
+            "patient.identifier must be in the format of "
+            f"\"{patient_identifier_system}|{{NHS number}}\" "
+            f"e.g. \"{patient_identifier_system}|9000000009\"")
+
+    if not nhs_number_mod11_check(patient_identifier_parts[1]):
+        raise ParameterException(f"Search parameter {patient_identifier_key} must be a valid NHS number.")
+
+    # immunization.target
```
dlzhry2nhs, line 61: Same for this, immunisation target processing can be abstracted into a standalone function.

backend/src/parameter_parser.py
```diff
@@ -36,6 +37,90 @@ def __repr__(self):
...
+    params[immunization_target_key] = list(set(params.get(immunization_target_key, [])))
+    vaccine_types_local = [v for v in params[immunization_target_key] if v is not None]
```
dlzhry2nhs, line 63: vaccine_types = {v for v in params.get(immunization_target_key, []) if v is not None}

Unless we need the side effect where we are reassigning `params[immunization_target_key]` (I really hope we do not), this should be done in one line.

Also it would be nice if we could change the variable to be a set instead. It expresses intent a bit better, but no worries if there are too many downstream changes to handle from changing the type.

backend/tests/test_parameter_parser.py
```diff
@@ -210,3 +212,83 @@ def test_create_query_string_with_multiple_immunization_targets_comma_separated(
...
+                    self.patient_identifier_key: [
+                        "https://fhir.nhs.uk/Id/nhs-number|1234567890"  # invalid mod11
+                    ],
+                    self.immunization_target_key: ["RSV"],
+                }
+            )
+
+        self.assertEqual(str(e.exception), f"Search parameter {self.patient_identifier_key} must be a valid NHS number.")
+
+    def test_process_search_params_invalid_include_value_is_rejected(self):
```
dlzhry2nhs, line 276: Apologies if there is a test outside of the diff, I am just going off the new tests.

Do we need a test case for when a the user provides an invalid vacc type in the list of vaccinations.

dlzhry2nhs: Also what is the expected behaviour when both NHS Number and the immunization target is invalid. Have you checked that Krupa has an acceptance criteria for this?

dlzhry2nhs (changes requested):
Couple of changes requested and queries about test cases.

Overall change looks good - in the right direction and the tests are well constructed.

JamesW1-NHS (dismissed):

backend/tests/test_parameter_parser.py
dlzhry2nhs: Yes that looks fine then. Can regard first comment as resolved...though I would like to see a test case where multiple are provided but 1 is invalid and multiple valid.

Point 2 will still need addressing however.

Akol125: That's already sorted with Krupa, validation errors for patient.identifier will be prioritized followed by imms.target

dlzhry2nhs (approved):

PR #867, opened 2025-10-02
VED-790-Redis-Schema-Storage

nhsdevws:
## Summary
* :sparkles: New Feature

Permit upload to redis cache of json schema. 
,
New version feathres:
-  permits upload of json files - with ".json" extension. 
- Unrecognized files will not be uploaded. Previously, unrecognised files result in exception. 
- json will be referenced with the filename as the key, in lowercase, and with an underscore instead of "." eg a.json gives key of "a_json"

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist


lambdas/redis_sync/src/transform_map.py
```diff
@@ -1,17 +1,19 @@
...
     logger.info("Transforming data for file type: %s", file_type)
     if file_type == RedisCacheKey.PERMISSIONS_CONFIG_FILE_KEY:
         return transform_supplier_permissions(data)
     if file_type == RedisCacheKey.DISEASE_MAPPING_FILE_KEY:
         return transform_vaccine_map(data)
+
     logger.info("No specific transformation defined for file type: %s", file_type)
-    return data  # Default case, return data as is if no transformation is defined
+
+    return transform_generic(data, file_type)
```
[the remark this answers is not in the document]
nhsdevws: Previously it also threw an exception. Unrecognized formats were not stored.

lambdas/redis_sync/src/transform_configs.py
```diff
@@ -46,3 +46,12 @@ def transform_supplier_permissions(mapping):
         "supplier_permissions": supplier_permissions,
         "ods_code_to_supplier": ods_code_to_supplier
     }
+
+
+def transform_generic(data, file_type) -> dict:
+    # check for generic json file
+    if file_type.lower().endswith('.json'):
```
dlzhry2nhs, line 53: If we're looking at validating the file extension, would it not be better to do this earlier before we've already started to read the file in the RedisCacher upload method?

lambdas/redis_sync/tests/test_transform_config.py
```diff
@@ -49,3 +57,40 @@ def test_empty_input(self):
             "supplier_permissions": {},
             "ods_code_to_supplier": {},
         })
+
+
+class TestTransformGeneric(TestBase):
```
dlzhry2nhs, line 62: Prefer not to have multiple classes in a test file, unless there is a good reason e.g. complex functionality such as different routes in the API/controller/service. Even then would favour splitting out into a different file.

Here, we are just testing some functions from a single file, so it should be part of the same class. If you are worried about the file reading done in setup for TestTransformConfigs then move them out into constants. They do not need to be freshly read on every iteration of a test.

dlzhry2nhs (changes requested):
Several points require addressing. Some minor, but a major issue concerning how file naming and key naming so that we will not have to rework this when it comes to components retrieving the validation schema.

lambdas/redis_sync/src/constants.py
```diff
@@ -1,3 +1,4 @@
 class RedisCacheKey:
     PERMISSIONS_CONFIG_FILE_KEY = "permissions_config.json"
     DISEASE_MAPPING_FILE_KEY = "disease_mapping.json"
+    VALIDATION_SCHEMA_FILE_KEY = "schema.json"
```
nhsdevws, line 4: filename should be ValidationRules.JSON

lambdas/redis_sync/tests/test_transform_config.py
```diff
@@ -2,13 +2,15 @@
...
 import json
 from unittest.mock import patch
-from transform_configs import transform_vaccine_map, transform_supplier_permissions
+from transform_configs import (
+    transform_vaccine_map, transform_supplier_permissions, transform_validation_schema
+)
 
 
-class TestTransformConfigs(unittest.TestCase):
+class TestBase(unittest.TestCase):
```
dlzhry2nhs, line 10: Could return it to original name I reckon.

dlzhry2nhs (approved):

PR #868, opened 2025-10-02
VED-821 Refactor get by ID journey in line with cleaner design

dlzhry2nhs:
## Summary
* Routine Change

Refactors the Get by ID journey to use the Controller -> Service -> Repository pattern properly. As this is quite a simple endpoint, I have also made quite a number of additional changes to show how I would like us to begin structuring the API code.

**Changes:**
- Added some Controller layer utilities for handling the parsing and response creation specific to API GW -> Lambda integration.
- Added top level exception wrapper to be used by controller methods. Means we can allow known custom exceptions to bubble up and return an appropriate response while also handling unexpected exceptions gracefully and getting a full stack trace.
- Vastly simplified get by ID journey to use the new utilities, raise errors where appropriate and ensure functionality is at the right level. (Due to the number of changes, it may be easiest to check out the branch and walk through the journey.)
- Create folders for controller, service, repository. Relevant functionality should go in those directories.
- Introduced a new utils directory for generic utilities across the project. I have just added a dictionary util for now. Changed utils in the testing folder to testing_utils to make the intent of this directory clearer, but also so it does not clash!

## Reviews Required
* [x] Dev

## Review Checklist

### DIFF
--- a/.github/workflows/deploy-backend.yml
+++ b/.github/workflows/deploy-backend.yml
@@ -56,7 +56,7 @@ jobs:
       contents: read
     steps:
       - name: Connect to AWS
-        uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a
+        uses: aws-actions/configure-aws-credentials@a03048d87541d1d9fcf2ecf528a4a65ba9bd7838
         with:
           aws-region: eu-west-2
           role-to-assume: arn:aws:iam::${{ vars.AWS_ACCOUNT_ID }}:role/auto-ops
--- a/.github/workflows/pr-teardown.yml
+++ b/.github/workflows/pr-teardown.yml
@@ -26,7 +26,7 @@ jobs:
   
     steps:
       - name: Connect to AWS
-        uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a
+        uses: aws-actions/configure-aws-credentials@a03048d87541d1d9fcf2ecf528a4a65ba9bd7838
         with:
           aws-region: eu-west-2
           role-to-assume: arn:aws:iam::${{ vars.AWS_ACCOUNT_ID }}:role/auto-ops
--- a/backend/poetry.lock
+++ b/backend/poetry.lock
@@ -27,18 +27,18 @@ files = [
 
 [[package]]
 name = "boto3"
-version = "1.40.42"
+version = "1.40.45"
 description = "The AWS SDK for Python"
 optional = false
 python-versions = ">=3.9"
 groups = ["main"]
 files = [
-    {file = "boto3-1.40.42-py3-none-any.whl", hash = "sha256:71ea8ca488e8b0d0a16b1b32816f8bab9aa66757c4122caf137ae2dbc95d7001"},
-    {file = "boto3-1.40.42.tar.gz", hash = "sha256:dbd8afc41f0cc759d0115c673e50bd1395b992a5c14627b56163cf35b86673c6"},
+    {file = "boto3-1.40.45-py3-none-any.whl", hash = "sha256:5b145752d20f29908e3cb8c823bee31c77e6bcf18787e570f36bbc545cc779ed"},
+    {file = "boto3-1.40.45.tar.gz", hash = "sha256:e8d794dc1f01729d93dc188c90cf63cd0d32df8818a82ac46e641f6ffcea615e"},
 ]
 
 [package.dependencies]
-botocore = ">=1.40.42,<1.41.0"
+botocore = ">=1.40.45,<1.41.0"
 jmespath = ">=0.7.1,<2.0.0"
 s3transfer = ">=0.14.0,<0.15.0"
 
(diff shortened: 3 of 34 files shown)


backend/src/controller/aws_apig_response_utils.py
```diff
@@ -0,0 +1,24 @@
+"""Utility module providing helper functions for dealing with response formats for AWS API Gateway"""
+import json
+from typing import Optional
+
+
+def create_response(
```
dlzhry2nhs, line 6: Moved from where it previously existed as a static method in the FHIR Controller. Could be made neater. Maybe even worth considering a class to abstract the concept of our API Responses in future.

backend/src/controller/fhir_controller.py
```diff
@@ -58,7 +57,7 @@ def get_immunization_by_identifier(self, aws_event) -> dict:
             else:
                 raise UnauthorizedError()
         except UnauthorizedError as unauthorized:
-            return self.create_response(403, unauthorized.to_operation_outcome())
+            return create_response(403, unauthorized.to_operation_outcome())
```
mfjarvis, line 60: This is a good example of us trapping exceptions unnecessarily. We'd get the same result if we let the `UnauthorizedError` reach the top level handler, but with less logic in the controller

mfjarvis: The get by ID looks so much cleaner 🙌

backend/src/repository/fhir_repository.py
```diff
@@ -102,19 +102,17 @@ def get_immunization_by_identifier(self, identifier_pk: str) -> tuple[Optional[d
         else:
             return None, None
 
-    def get_immunization_by_id(self, imms_id: str) -> Optional[dict]:
+    def get_immunization_by_id(self, imms_id: str) -> tuple[Optional[dict], Optional[str]]:
```
mfjarvis, line 105: I do think `tuple` is a less misleading return value than the previous `dict`, but I still had to read the last line to know what that tuple contained. Maybe something like `get_immunisation_and_version_by_id` would be clearer?

backend/tests/controller/test_fhir_api_exception_handler.py
```diff
@@ -0,0 +1,75 @@
...
+
+        for msg, error_type, expected_status, expected_code, expected_message in test_cases:
+            with self.subTest(msg=msg):
+
+                @fhir_api_exception_handler
+                def dummy_func():
+                    if msg == "Test ResourceNotFoundError":
+                        raise error_type(resource_type="Immunization", resource_id="123")
+
+                    raise error_type()
```
mfjarvis, line 44: Could provide the exact error to raise in the test case, then we wouldn't need the `msg` logic:
```suggestion
        test_cases = {
            (UnauthorizedError(), 403, "forbidden", "Unauthorized request"),
            (UnauthorizedVaxError(), 403, "forbidden", "Unauthorized request for vaccine type"),
            (
                ResourceNotFoundError(resource_type="Immunization", resource_id="123"),
                404,
                "not-found",
                "Immunization resource does not exist. ID: 123"
            )
        }

        for error, expected_status, expected_code, expected_message in test_cases:
            with self.subTest(msg=f`Test {error.__class__.__name__}`):

                @fhir_api_exception_handler
                def dummy_func():
                    raise error()
```

backend/src/utils/dict_utils.py
```diff
@@ -0,0 +1,21 @@
...
+    if not target_dict or not isinstance(target_dict, dict):
+        return default
+
+    latest_nested_dict = dict(target_dict)
+
+    for key in args:
+        if key not in latest_nested_dict:
+            return default
+
+        if not isinstance(latest_nested_dict[key], dict):
```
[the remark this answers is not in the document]
mfjarvis: Ah no, the `key not in` check will catch that. So is this line needed?

backend/src/controller/aws_apig_event_utils.py
```diff
@@ -0,0 +1,28 @@
+"""Utility module for interacting with the AWS API Gateway event provided to controllers"""
+from typing import Optional
+
+from aws_lambda_typing.events import APIGatewayProxyEventV1
```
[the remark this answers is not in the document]
dlzhry2nhs: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html From this I was 80% sure given that we get multivalue headers in the payload.

I then looked in a bit more depth and found that `oas.yaml` is where our config is hidden, and we do indeed use payload version 1. I can change the search endpoint when it comes to refactoring it.

backend/tests/controller/test_fhir_api_exception_handler.py
dlzhry2nhs: Done - much nicer.

backend/src/controller/aws_apig_event_utils.py
```diff
@@ -0,0 +1,28 @@
...
+
+from controller.constants import SUPPLIER_SYSTEM_HEADER_NAME
+from models.errors import UnauthorizedError
+from utils import dict_utils
+
+
+def get_path_parameter(event: APIGatewayProxyEventV1, param_name: str) -> str:
+    return dict_utils.get_field(
+        dict(event),
+        "pathParameters",
```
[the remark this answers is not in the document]
dlzhry2nhs: Done

backend/src/service/fhir_service.py
```diff
@@ -83,25 +83,21 @@ def get_immunization_by_identifier(
         imms_resp['resource'] = filtered_resource
         return form_json(imms_resp, element, identifier, base_url)
 
-    def get_immunization_by_id(self, imms_id: str, supplier_system: str) -> Optional[dict]:
+    def get_immunization_by_id(self, imms_id: str, supplier_system: str) -> tuple[Immunization, str]:
```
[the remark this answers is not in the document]
dlzhry2nhs: Done for both.

backend/src/controller/aws_apig_event_utils.py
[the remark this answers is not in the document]
mfjarvis: You're right of course! I thought it was a simple correspondence between REST API: V1 and HTTP API: V2.

PR #882, opened 2025-10-07
VED-814 Resolve sonarcloud warnings on S3 bucket ownership

dlzhry2nhs:
## Summary
* Routine Change

Resolves the outstanding security warnings we have in Sonarcloud that causes us to fail on `master`/overall coverage: https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&types=VULNERABILITY&severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&inNewCodePeriod=true&sinceLeakPeriod=true&id=NHSDigital_immunisation-fhir-api

- Add expected bucket owner account to boto3 S3 calls
- Update tests to pass using the moto default account ID

## Reviews Required
* [ ] Dev

## Review Checklist

### DIFF
--- a/mesh_processor/src/converter.py
+++ b/mesh_processor/src/converter.py
@@ -5,6 +5,7 @@ from typing import BinaryIO
 import boto3
 from smart_open import open
 
+EXPECTED_BUCKET_OWNER_ACCOUNT = os.getenv("ACCOUNT_ID")
 DESTINATION_BUCKET_NAME = os.getenv("DESTINATION_BUCKET_NAME")
 
 logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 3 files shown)


mesh_processor/src/converter.py
```diff
@@ -80,9 +81,15 @@ def move_file(source_bucket: str, source_key: str, destination_bucket: str, dest
     s3_client.copy_object(
         CopySource={"Bucket": source_bucket, "Key": source_key},
         Bucket=destination_bucket,
-        Key=destination_key
+        Key=destination_key,
+        ExpectedBucketOwner=EXPECTED_BUCKET_OWNER_ACCOUNT,
+        ExpectedSourceBucketOwner=EXPECTED_BUCKET_OWNER_ACCOUNT
```
Akol125, line 86: just curious as to why two arguments references the same value

mfjarvis (approved):

Akol125 (approved):

mesh_processor/src/converter.py
Akol125: Hence the same account ID, makes total sense

PR #1382, opened 2026-04-07
VED-1157: Ack backend Lambda deployment to a dedicated pipeline stage

Thomas-Boyle:
- Added support for building and deploying the ack backend image in the continuous deployment, backend deployment, and PR workflows.
- Introduced new variables for ack backend image version and build options in the relevant YAML files.
- Created a new ECR repository for the ack backend and updated the Terraform configuration accordingly.
- Enhanced the teardown process to clean up the ack backend image references.

This update ensures that the ack backend is properly integrated into the CI/CD pipeline, improving deployment consistency and management.


.github/workflows/deploy-backend.yml
```diff
@@ -122,18 +143,43 @@ jobs:
...
+
   terraform-plan:
     permissions:
       id-token: write
       contents: read
     needs:
       - deploy-recordprocessor-image
-    if: ${{ !cancelled() && needs.deploy-recordprocessor-image.result == 'success' }}
+      - deploy-ack-backend-image
+    if: ${{ !cancelled() && needs.deploy-recordprocessor-image.result == 'success' && needs.deploy-ack-backend-image.result == 'success' }}
```
avshetty1980: should we use `!= 'failure' ` instead of `== 'success' ` ? this way it wouldnt block if deploy-lambda-artifact.yml (should we change the filename?) ever completes as `skipped`.

infrastructure/instance/ack_lambda.tf
```diff
@@ -1,81 +1,5 @@
...
-        ],
-        "Condition" : {
-          "StringLike" : {
-            "aws:sourceArn" : "arn:aws:lambda:${var.aws_region}:${var.immunisation_account_id}:function:${local.ack_lambda_name}"
-          }
-        }
-      }
-    ]
-  })
+  ack_lambda_name = "${local.short_prefix}-ack-lambda"
```
avshetty1980, line 2: do we need a removed block for `aws_ecr_repository.ack_lambda_repository` because terraform destroy might try to destroy it but the plan will not include it? eg:
```
removed {
  from = aws_ecr_repository.ack_lambda_repository
  lifecycle { destroy = false }
}
```

.github/workflows/deploy-backend.yml
Thomas-Boyle: I think keep `== 'success'` here because terraform plan consumes `image_uri` outputs from both reusable workflow jobs. If either upstream job is skipped, those outputs won’t be guaranteed, so `!= 'failure'` could unblock plan with empty image references.

infrastructure/instance/ack_lambda.tf
Thomas-Boyle: These are now added. Good shout!

.github/workflows/deploy-backend.yml
```diff
@@ -122,18 +143,43 @@ jobs:
...
     outputs:
       recordprocessor_image_uri: ${{ needs.deploy-recordprocessor-image.outputs.image_uri }}
+      ack_backend_image_uri: ${{ needs.deploy-ack-backend-image.outputs.image_uri }}
     runs-on: ubuntu-latest
```
avshetty1980, line 163: do we need an output for ack_backend_image_uri in outputs.tf? eg:
```
output "ack_backend_image_uri" {
  description = "Immutable URI of the deployed ack backend Lambda image"
  value       = var.ack_backend_image_uri
}
```

.github/workflows/pr-deploy-and-test.yml
```diff
@@ -21,6 +21,7 @@ jobs:
     with:
       apigee_environment: internal-dev
       build_recordprocessor_image: ${{ github.event.action == 'opened' || github.event.action == 'reopened' }}
+      build_ack_backend_image: ${{ github.event.action == 'opened' || github.event.action == 'reopened' }}
       diff_base_sha: ${{ github.event.before }}
       diff_head_sha: ${{ github.event.pull_request.head.sha }}
       run_diff_check: ${{ github.event.action == 'synchronize' }}
```
avshetty1980, line 29: will a force push reset the `diff_base_sha `to the wrong sha?

.github/workflows/deploy-backend.yml
Thomas-Boyle: I don’t think we need an output here. In this flow `ack_backend_image_uri` is only an input into the instance stack, and the GitHub Actions job output is already being passed through directly from `deploy-ack-backend-image`. Adding to `outputs.tf` would only echo the input variable unless we have another consumer reading it from Terraform state.

avshetty1980: fair. 
when we have lots of lambdas moving to this pattern, would it be good to have it output in the CI logs? your call - im easy here

.github/workflows/pr-deploy-and-test.yml
Thomas-Boyle: After checking the logic I think we can keep existing setup. A force push could make `github.event.before` point to an old head and widen diff, but that would only cause an unnecessary rebuild, not miss a required one. In this case I’d prefer that safe fallback over changing diff semantics.

.github/workflows/deploy-backend.yml
```diff
@@ -14,6 +14,14 @@ on:
         required: false
         type: string
         default: ""
+      build_ack_backend_image:
```
avshetty1980: I was thinking of a repeatable pattern and introduce a matrix pattern strategy. Something like this:
```
lambda_image_overrides:
  description: JSON map of {lambda_name: image_uri} for reuse mode. Empty = build all.
  required: false
  type: string
  default: "{}"

# Single job replacing all per-lambda deploy jobs:
deploy-lambda-images:
  strategy:
    matrix:
      include:
        - lambda_name: recordprocessor
          ecr_repository: imms-recordprocessor-repo
          dockerfile_path: lambdas/recordprocessor/Dockerfile
          lambda_paths: lambdas/recordprocessor/
        - lambda_name: ack-backend
          ecr_repository: imms-ackbackend-repo
          dockerfile_path: lambdas/ack_backend/Dockerfile
          lambda_paths: lambdas/ack_backend/
        # ... ......
  uses: ./.github/workflows/deploy-lambda-artifact.yml
  with:
    lambda_name: ${{ matrix.lambda_name }}
    image_version: ${{ fromJson(inputs.lambda_image_overrides)[matrix.lambda_name] || '' }}
# .....
```
We can then pass all image URIs to terraform as a single JSNO map var or maybe even use `jq` built output step that assembles the TF_VAR_* env vars. This way adding a new lambda will just be adding it to the matrix entry ...
thoughts

.github/workflows/pr-deploy-and-test.yml
avshetty1980: cool.
maybe test this in the PR too, just to be sure.

.github/workflows/deploy-backend.yml
Thomas-Boyle: So we'd be looking at a pattern sort of like: matrix deploy -> collect uploaded manifest artifacts -> build single JSON map of `{lambda_name: image_uri}` -> feed Terraform from that.

Is that change something that would be better doing as a separate ticket once the all the lambda's have been refactored?

avshetty1980: I would recommend making the change now so its not another ticket for the refactor and the effort required in the rest of the lambda updates would be smaller, hopefully.
I think `deploy-lambda-artifact.yml` already uploads a manifest/lambda and actions/download-artifact@v4 supports `pattern: `with `merge-multiple:true`

https://oneuptime.com/blog/post/2025-12-20-artifacts-github-actions/view

 in terraform-plan we can:
- do the download all lambda manifests in a single step
-  then, set TF_VAR env var from manifests (Also emit the raw map so terraform-apply can reconstruct it without re-downloading artifacts (they may expire between jobs)

then in terraform-apply, we can:

- restore the TF_VAR env vars from plan outputs
- ...

avshetty1980: if you use the above process then we might also have to update the deploy-backend.yml inputs which have image promotion but with matrix we wont be able to pass per-lambda version overrides unless we change to a single JSON map:
in the inputs we can have:
```
lambda_image_overrides:
  description: >
    JSON map of lambda_name -> image_version for reuse mode.
    e.g. {"recordprocessor": "internal-dev-git-abc123", "ack-backend": "..."}
  required: false
  type: string
  default: "{}"
```
then in the matrix job:
`image_version: ${{ fromJSON(inputs.lambda_image_overrides)[matrix.lambda_name] || '' }}`
and tge continuous-deployment.yml would just pass:
```
lambda_image_overrides: >-
  ${{ toJSON(needs.deploy-internal-dev-backend.outputs.image_uris_json) }}
```
where the image_uris_json is the output from the internal-dev run

Thomas-Boyle: This is now implemented.

.github/workflows/deploy-backend.yml
```diff
@@ -127,13 +155,11 @@ jobs:
       id-token: write
       contents: read
     needs:
-      - deploy-recordprocessor-image
-    if: ${{ !cancelled() && needs.deploy-recordprocessor-image.result == 'success' }}
+      - deploy-lambda-images
+    if: ${{ !cancelled() }}
```
avshetty1980: with `fail-fast:false` does this mean that if the ack-backend build fails then terraform-plan will still run?
also, if terraform-plan passes with a "placeholder" will terraform-apply run?

.github/workflows/deploy-backend.yml
```diff
@@ -100,25 +115,38 @@ env: # Sonarcloud - do not allow direct usage of untrusted data
...
+              lambdas/ack_backend/
     uses: ./.github/workflows/deploy-lambda-artifact.yml
     with:
-      lambda_name: recordprocessor
+      lambda_name: ${{ matrix.lambda_name }}
       environment: ${{ inputs.environment }}
       sub_environment: ${{ inputs.sub_environment }}
-      build_image: ${{ inputs.build_recordprocessor_image }}
-      image_version: ${{ inputs.recordprocessor_image_version }}
+      build_image: ${{ matrix.lambda_name == 'recordprocessor' && inputs.build_recordprocessor_image || matrix.lambda_name == 'ack-backend' && inputs.build_ack_backend_image || false }}
```
avshetty1980: instead of this pattern - which could start getting a bit unwieldy when we add more lambdas, is it worth using a `lambda_build_flags` in the inputs of `workflow_dispatch` like so:
```
lambda_build_flags:
  description: 'JSON map of lambda_name -> bool. e.g. {"recordprocessor":true,"ack-backend":false}'
  required: false
  type: string
  default: "{}"
```

then in the matrix `with` section we could do something like:
`build_image: ${{ fromJson(inputs.lambda_build_flags)[matrix.lambda_name] || false }}`

and then callers like pr-deploy-and-test.yml can just use:
```
lambda_build_flags: >-
  ${{ (github.event.action == 'opened' || github.event.action == 'reopened')
      && '{"recordprocessor":true,"ack-backend":true}'
      || '{}' }}
```

infrastructure/account/ackbackend_ecr_repo.tf
```diff
@@ -0,0 +1,53 @@
...
+        Effect = "Allow"
+        Principal = {
+          Service = "lambda.amazonaws.com"
+        }
+        Action = [
+          "ecr:BatchGetImage",
+          "ecr:DeleteRepositoryPolicy",
+          "ecr:GetDownloadUrlForLayer",
+          "ecr:GetRepositoryPolicy",
+          "ecr:SetRepositoryPolicy"
```
avshetty1980: does the lambda service principal need `DeleteRepositoryPolicy`, `GetRepositoryPolicy` and `SetRepositoryPolicy` ? if not, then I think we can also amend this in `ack_lambda_ECRImageRetreival_policy` and `recordprocessor`

.github/workflows/deploy-backend.yml
```diff
@@ -100,25 +115,38 @@ env: # Sonarcloud - do not allow direct usage of untrusted data
 run-name: Deploy Backend - ${{ inputs.environment }} ${{ inputs.sub_environment }}
 
 jobs:
-  deploy-recordprocessor-image:
-    name: Deploy recordprocessor image
+  deploy-lambda-images:
+    name: Deploy ${{ matrix.lambda_name }} image
+    strategy:
+      fail-fast: false
```
avshetty1980, line 118: is this so we can see all the failures in one run? is this better than failing fast true?

avshetty1980 (commented):
I think we're nearly there - just a few comments and nits.

infrastructure/account/ackbackend_ecr_repo.tf
Thomas-Boyle: I don't think it needs these, I've removed them.

.github/workflows/deploy-backend.yml
Thomas-Boyle: An ack_backend image failure will not let terraform-plan run, even with fail-fast false. fail-fast false only affects whether the other matrix leg is cancelled early.
Terraform-apply will not run unless terraform-plan itself completed successfully. There is no placeholder success path here, if manifests are missing the plan job exits non zero before Terraform runs

.github/workflows/deploy-backend.yml
Thomas-Boyle: Yep, the intent of fail-fast: false here is to let both matrix legs run so we can see all lambda image failures in one run, rather than stopping at the first one. In this workflow that felt preferable because the lambda builds are independent.

.github/workflows/deploy-backend.yml
Thomas-Boyle: Done.

.github/workflows/deploy-backend.yml
```diff
@@ -151,6 +174,38 @@ jobs:
...
+          )"
+
+          echo "image_uris_json=${image_uris_json}" >> "$GITHUB_OUTPUT"
+          jq -er '
+            to_entries[]
+            | "TF_VAR_\(.key | gsub("-"; "_"))_image_uri=\(.value)"
+          ' <<< "${image_uris_json}" >> "$GITHUB_ENV"
+
       - name: Terraform Init
         working-directory: infrastructure/instance
```
avshetty1980, line 234: if we used this pattern, will id-sync lambda produce `TF_VAR_id_sync_image_uri` and so is there any enforcement to ensure that this is what the variable name chosen is too?
would it be better to add it in the matrix definition, like so:
```
matrix:
  include:
    - lambda_name: recordprocessor
      tf_var_suffix: recordprocessor   # TF_VAR_recordprocessor_image_uri
      ecr_repository: imms-recordprocessor-repo
      ...
    - lambda_name: ack-backend
      tf_var_suffix: ack_backend       # TF_VAR_ack_backend_image_uri
      ecr_repository: imms-ackbackend-repo
      ...
```

Thomas-Boyle: Agreed, now added.

avshetty1980 (approved):
Nice! really good PR that we can reuse over all our remaining lambdas 💯

avshetty1980 (dismissed):

avshetty1980 (approved):

amarauzoma (approved):

